PR #16, opened 2015-03-09
Implement tasklist

mizchi:
This is a feature of gfm.

[Task Lists in GFM: Issues/Pulls, Comments](https://github.com/blog/1375%0A-task-lists-in-gfm-issues-pulls-comments "Task Lists in GFM: Issues/Pulls, Comments")

I had to implement for my works, but it's not in commonmark specs.  I'll leave it up to you.

(And I feel a bit dirty implementation)

## Exmaple

```javascript
var mdast = require('./index');
var util = require('util');
var parsed = mdast.parse('- [x] a\n- [ ] b', {tasklist: true});
console.log(util.inspect(parsed, false, null));
```

result

```javascript
{ type: 'root',
  children: 
   [ { type: 'list',
       ordered: false,
       children: 
        [ { type: 'listItem',
            loose: false,
            children: 
             [ { type: 'paragraph',
                 children: 
                  [ { type: 'text',
                      value: '[x] a',
                      position: { start: { line: 1, column: 3 }, end: { line: 1, column: 8 } } } ],
                 position: { start: { line: 1, column: 3 }, end: { line: 1, column: 8 } } } ],
            checked: true,
            position: { start: { line: 1, column: 1 }, end: { line: 1, column: 8 } } },
          { type: 'listItem',
            loose: false,
            children: 
             [ { type: 'paragraph',
                 children: 
                  [ { type: 'text',
                      value: '[ ] b',
                      position: { start: { line: 2, column: 3 }, end: { line: 2, column: 8 } } } ],
                 position: { start: { line: 2, column: 3 }, end: { line: 2, column: 8 } } } ],
            checked: false,
            position: { start: { line: 2, column: 1 }, end: { line: 2, column: 8 } } } ],
       position: { start: { line: 1, column: 1 }, end: { line: 2, column: 8 } } } ],
  position: { start: { line: 1, column: 1 }, end: { line: 2, column: 8 } } }
```

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -984,14 +984,35 @@ LIST_ITEM_MAP.false = renderNormalListItem;
  * @return {Object}
  */
 function renderListItem(token, position) {
-    token = LIST_ITEM_MAP[this.options.pedantic].apply(this, arguments);
+    var item;
+    var checked;
 
-    return {
+    // detect checkbox and consume as checked
+    if (this.options.tasklist) {
+      if(/^-\s\[x\]/.test(token)) {
+        checked = true;
+        token = token.replace(/^\s\[x\]/, '');
+      }
+      else if(/^-\s\[\s\]/.test(token)) {
+        checked = false;
+        token = token.replace(/^\s\[\s\]/, '');
+      }
+    }
+
+    token = LIST_ITEM_MAP[this.options.pedantic].call(this, token, position);
+
+    item = {
         'type': LIST_ITEM,
         'loose': EXPRESSION_LOOSE_LIST_ITEM.test(token) ||
             token.charAt(token.length - 1) === NEW_LINE,
         'children': this.tokenizeBlock(token, position)
     };
+
+    if (checked != null) {
+      item.checked = checked
+    }
+
+    return item;
 }
 
 /**


wooorm:
Hey! Thanks for the PR. I have to say that I’d like to rework this a bit though—add docs, support for other list item markers (`*`, `+`), tests, enable this through the `gfm` setting instead of its own option.

I’m currently in the process of a rewrite for the test proces—I’ll probably have this functionality added somewhere this week!

wooorm:
I just finished the test-rewrite, and will be working on this next. I expect it working in one or two days.

wooorm:
I hope you like the addition!

-    This update supports asterisks, plusses, and numbered list items;
-    This update eats the checkboxes instead of letting them stay in a `text` node;
-    This update does not support new lines for unchecked checkboxes (which GFM supports for, if i remember correctly, only for ordered lists);
-     GFM also has some weird differences between which marker is used, and whether the list is ordered or not, whereas this addition has clear support between these differences.

mizchi:
Thx for your guide!

This weekend and next week may be busy. Anyway I will do it.

wooorm:
Do what? Implement it? Because I already [did that](https://github.com/wooorm/mdast/commit/ec9c2cb15705814a2cd36e9bd09a4d84048afb74)—it’s released in mdast@0.11!

mizchi:
Oh I misundertood! You did? Great!

PR #38, opened 2015-07-14
Fix order of multiple attachers in mdast.use

### LINKED ISSUE #37: Inverse order of attachers when passed in array
Looking at [index.js](https://github.com/wooorm/mdast/blob/f693521ddb42e2ae581cdf6825bbc7058e51b017/index.js#L158-160), it seems that

```
mdast.use([plugin1, plugin2, plugin3])
```

is equivalent to

```
mdast.use(plugin3).use(plugin2).use(plugin1)
```

which is counter-intuitive if you ask me.

Why is it so?

eush77:
This patch changes the order in which multiple attachers (passed in array) are attached in `mdast.use`.

```
mdast.use([plugin1, plugin2, plugin3])
```

becomes equivalent to

```
mdast.use(plugin1).use(plugin2).use(plugin3)
```

Close #37.

### DIFF
--- a/index.js
+++ b/index.js
@@ -153,9 +153,9 @@ function use(attach, options) {
      */
 
     if ('length' in attach && typeof attach !== 'function') {
-        index = attach.length;
+        index = -1;
 
-        while (attach[--index]) {
+        while (attach[++index]) {
             self.use(attach[index]);
         }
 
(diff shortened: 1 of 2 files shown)


wooorm:
Thanks! :+1:

PR #50, opened 2015-09-10
Add mdast-autolink-headings to the plugin list.

ben-eb:


wooorm:
Thanks! Could you PR on mdast-html too?

ben-eb:
Sure.

PR #53, opened 2015-09-12
Add mention of use in Zotero citations

retorquere:


wooorm:
Thanks! However, I decline because the list of plugins is not the correct place for this (it doesn’t use the plug-in system mdast has). This tool is “built using mdast”, which should be a separate page. I’ll leave this open until I find a nice place for this.

retorquere:
Agreed that it's not a general-use plugin (although it does use the plugin system: https://github.com/ZotPlus/zotero-citations/blob/master/lib/zotero-citations.coffee#L12)

wooorm:
In that case you should externalise `schomd` as a separate plugin! Take a look at the other plugins for how the READMEs are layed-out. A nice example is https://github.com/ben-eb/mdast-midas.

retorquere:
I don't have a good plan in place to make this a more general schmod processor -- it's thoroughly dependent on Better BibTeX running right now. Should I wait for that, or publish as-is?

wooorm:
Wait for that IMHO if that’s what makes it useful!

PR #56, opened 2015-09-13
Mention mdast's use in github-man, npm-man

eush77:


wooorm:
:+1:

wooorm:
Weird that the build fails on v3.3.0, I tested it locally and nothing’s wrong.

PR #57, opened 2015-09-14
Implement --watch

chcokr:
This commit passes `npm run lint` and `npm test`.

I will need some guidance with documentation and tests.
- Documentation: Can you point me to the files I will need to update?
- Tests: I'm not sure how to a watch feature.

Thanks in advance for your feedback.


lib/cli/file-pipeline/file-system.js
```diff
@@ -14,6 +14,7 @@
 
 var fs = require('fs');
 var debug = require('debug')('mdast:cli:file-pipeline:file-system');
+var watchOutputCache = require('../watchOutputCache');
```
chcokr: Probably should rename the file to `../watch-output-cache`

wooorm:
Thanks a lot! Really appreciate the work! I was offline so I couldn’t add line-notes:

Okey, here comes a long list of “issues”. Don’t let them discourage you though,
I like the work a lot, I just have very strict code-standards.

Overall:

*   Rebase with master :smile:

*   Bug: `mdast --watch` fails without warning, should be treated like `mdast`
    (by failing).

*   I tested it a bit, works nice! Main point of interest I’m missing some some
    or progress meter: Something like `sindresorhus/elegant-spinner` would be
    wonderful.

*   Style: I love blank-lines between things (e.g., `cli/index.js:55-56`), but
    not at the edges of blocks (`cli/index.js:46` and `cli/index.js:64`), could
    you refactor to match my style? Sorry to be nitpicking.

*   P.S. I can help with programming too, if you’d like.

cli/cli.js:

*   Could you place `-w, --watch` a bit higher? Below `-e, --ext` is fine;

*   Also: how about `watch for changes and reprocess (ignores output)` as a
    description?

cli/file-pipeline/file-system.js

*   78: `watched` is just as nice a variable name;

*   78: Checks if files are output, but not if they’re output to a watched file,
    use `context.fileSet.sourcePaths` and `file.sourcePath` here.

*   91: Instead of emitting a warning on each file and every run, which can
    get huge, how about emitting a warning just once, at the top, and only on
    first run? E.g., a warning could be emitted on `cli/index.js:46`.

*   91: Could you add some line-breaks? E.g.,
    `file.warn(['...', '...'].join('\n'))`

watchOutputCache.js

*   Switch casing to dash-case;
*   Could you add a jsdoc comment at the top? Just like the other files.
*   13: Use `file.filePath()`;
*   36: Use a block comment here: `/*\n *\n */`

cli/index.js:

*   48: You can use `fileSet.sourcePaths` here, I think it’s available on `cli`;

*   53: Can watcher be chained? If so, please do :)

*   58: Could we make sure not every file is reprocessed? It’s “nice to have”
    and not essential, but it would be very cool.

wooorm:
Other than that, again, really :+1:

chcokr:
> cli/index.js 48: You can use fileSet.sourcePaths here, I think it’s available on cli;

I took a look, and `cli.fileSet.sourcePaths` doesn't seem to return what I want. It seems to return a list of all files that will be affected after the plugins are run. That's not what I want here - I want a list of files that are being watched (the arguments in `argv`). I'll leave a comment there.

chcokr:
> cli/file-pipeline/file-system.js 78: Checks if files are output, but not if they’re output to a watched file,
use context.fileSet.sourcePaths and file.sourcePath here.

Same concern from my above comment. Bu yeah, I will fix `not if they’re output to a watched file`

wooorm:
Great work, thanks so much for taking the time :+1:

chcokr:
Thanks!

> a warning could be emitted on cli/index.js:46

What's the best way to print a non-file-specific warning? Something like `console.warn()`?

wooorm:
@chcokr Hmm, there’s no support for warning on the file-set yet, maybe `cli.stderr()` can be used?

wooorm:
@chcokr Are you sure `sourcePaths` isn’t usable? It should be the same as [`cli.files`](https://github.com/wooorm/mdast/blob/8c80af99996e04571f1d3818a5c68041eb929711/lib/cli/file-set-pipeline/transform.js#L32) initially, but it [prevents](https://github.com/wooorm/mdast/blob/8c80af99996e04571f1d3818a5c68041eb929711/lib/cli/file-set.js#L176-L182) changing files from being re-added and duplicates?

chcokr:
I'm pretty sure. Add the following two lines at line 68 in `cli/index.js` and see the differences between the two arrays on initial startup.

```JS
console.log(cli.fileSet.sourcePaths);
console.log(fileAbsolutePaths);
```

Also, I think it's better to pass **absolute** paths into chokidar, as I explained in commit 008b3fa just now.

chcokr:
> progress meter

Exactly at what stage would you want this meter to be placed? Between every `fileSetPipeline.run()`? I'm not exactly clear on what kind of visualization you're expecting.

wooorm:
@chcokr I get exactly the same results when I add your lines (but pass all `resolve` the sourcePaths first though)

```js
cli.fileSet.sourcePaths.map(function (f) {
  return path.resolve(f);
})
```

chcokr:
My point is this (assuming you do `mdast -w readme.md`):
- `fileSet.sourcePaths` returns 4 files
- but what I want here is only 1 file, because `chokidar` should be watching just `readme.md`, not the other three

wooorm:
P.S. I did `bin/mdast . -w` in the repo.

A great example is [speed-test](https://github.com/sindresorhus/speed-test), that’s just too beautiful. I’m not sure what some useful messages would be though. However, I did three saves on the screenshot below, the first without errors, the second with one, the third again without errors. But it’s not very obvious (I did `mdast . -wq`, which is more beautful / minimal)!

screen shot 2015-09-14 at 12 27 03 p m

wooorm:
@chcokr Oh you’re right, sorry, plug-ins can add files to the file-set too, but they will never be written / logged about (that’s the `providedByUser` variable you see in some places)... Hmmm.

However, should those files be watched? Maybe yes, it’s used by `mdast-validate-links` and I can imagine that being useful!

chcokr:
I think you posted a completely irrelevant screenshot?

chcokr:
Ah, you fixed the screenshot :P

wooorm:
Yeah, sorry :/

chcokr:
I'll try to come up with a reasonable spinner idea. But let's get the `fileSet.sourcePaths` thing figured out.

I don't think you should watch the three other files. Because, let's think from the user's perspective. When the user says `mdast -wq readme.md`, he expects just `readme.md` to be watched, not the other three files. Wouldn't you agree that watching the three other files doesn't align with the user's expectations?

wooorm:
Yeah the spinner should be an enhancement :)

Re your point: I’m not sure, the use case here is that the user wan’t to validate links, so if he changes one of the dependent files, why not reload?

Maybe it’s not such an important use-case though, if watching just `readme` is indeed the goal?
We could also just check for `fileSet.valueOf()....sourcePath`, instead of `fileSet.sourcePaths`, I still like that because it ensures generated files like man pages **can** get reloaded safely.

wooorm:
Point of interest re my motivation, I’d like to externalise the CLI in the future for other tools, like wooorm/retext and wooorm/hast, and the more to come processors: tools which might by default almost always generate new files instead of regenerate source files.

wooorm:
Also: right now if I’m using the `npm run build-man` script with watch, the wrong files are watched (the man pages, not the docs), because the location of the file is changed. At least resolve `file.sourcePath || file.filePath()` on files instead of just checking for `file.filePath()`, that should be good enough I think?

wooorm:
Just found out mdast-man had a bug, [should be fixed](https://github.com/wooorm/mdast-man/commit/101df38d30ac78fe49e66a94b306e5dfc541905a), so the `npm run build-man` should work to test!

chcokr:
After testing with `mast-validate-links`, `npm run build-man` (thanks for the fix), etc., I have come to the conclusion that it is best to just use `cli.fileSet.sourcePaths`. My reasoning is explained in the commit e7717cb.

wooorm:
Cool! Were you able to run the man task with watch and indeed have it update the man pages on changes?

chcokr:
Mmhm! Just don't forget you have to wait until you press CTRL+C

chcokr:
Alright, working on the spinner part now.

I want to output a message when there is no output after a pipeline run. Where in the codebase should I do this?

wooorm:
@chcokr Why? There’s no reason to stop writing to non-watched files right? Maybe if [L80](https://github.com/chcokr/mdast/blob/watch/lib/cli/file-pipeline/file-system.js#L80) here is changed to `destinationPath ` that can be ignored? Because, the file being written to is not watched?

wooorm:
Reporting is here: https://github.com/wooorm/mdast/blob/master/lib/cli/file-set-pipeline/log.js

chcokr:
I'm not sure what you're talking about - are you rebutting my "output a message when there is no output"?

wooorm:
My first response was for [this](https://github.com/wooorm/mdast/pull/57#issuecomment-140415617) comment, my second for your [other one](https://github.com/wooorm/mdast/pull/57#issuecomment-140416448) :smile:

chcokr:
Ah OK. Your first response makes sense. I have to leave, so I'll be back in a couple hours.

Also, now that I think about "output a message when there is no output" again, I actually don't think that's necessary. The spinner will have stopped, and I think that will be enough indication that there has been no output.

wooorm:
:+1: Yeah, we could also just make an options for the reporter to output something when no messages are found, similar to the `quiet`, `silent`, options

wooorm:
I just added the 'destinationPath' discussed above privately and it made the man task work :)

chcokr:
I'm not sure what you mean by "man task work"ing. Can you show me the code you had to add to make the program write to non-watched files without waiting until SIGINT occurs?

chcokr:
If that's not what you mean... my code as of 290b7a7 already had the man task work, though you had to wait until SIGINT.

wooorm:
I was referring to this change: https://github.com/wooorm/mdast/pull/57#issuecomment-140416657

chcokr:
Right, but I don't see how changing `file.sourcePath` to `destinationPath` on that line changes anything. Aren't they almost identical?

wooorm:
Nope, files move: e.g., when `mdast-man` is used, `mdast.1.md` moves to `mdast.1`, same goes for when `--ouput man/` is used to move files to another directory. Sorry I wasn’t more clear, but that’s a big use case for me, and there’s no reason to *not* write files to unwatched paths when watching!

chcokr:
Ah OK, I'll take a look later. Right now I'm working on the spinner.

I found a new issue: If I run `bin/mdast doc/*.?.md -c .mdastrc-man -q -w` with my latest commit, and then deliberate make a lint error on, say, `mdast.1.md`, I get multiple warnings of the same kind. Do you now why this happens?

wooorm:
I just tested that and didn’t get a warning (I added a newline between the first and its following header), with your exact script

wooorm:
a second warning*

screen shot 2015-09-14 at 8 29 16 p m

chcokr:
Hmm, do this: make any line longer past 80 characters.

wooorm:
Still one warning: I’m on chcokr/mdast@4f7e2b1

wooorm:
Whoops, I mean chcokr/mdast@290b7a7, the last commit is because i had to rebase with master, that shouldn’t be it but maybe?

chcokr:
Hmm. This is what happens when I'm on 290b7a7:

screen shot 2015-09-15 at 2 51 02 pm

wooorm:
Could you `git checkout` previous commits and see where I started? I might also be your system? What OS are you on?

chcokr:
> see where I started

I'm not sure what you mean by this?

I'm on OS X Yosemite.

wooorm:
it* So you can checkout the previous commit and try if that fails, then before that, &c.

Also: I’m on OS X Yosemite too :/

chcokr:
What the heck, I'm not getting it any more either on 290b7a7. Never mind, I'm forgetting about it until it comes back up

chcokr:
Hey, I just came across another new issue on 290b7a7.

Do `mdast -wq readme.md`. While the watch is running, change the content of line 106 on `readme.md`. Hit CTRL+C so that the output cache is released. Does the content of the block at line 100 change for you?

wooorm:
Yup, seems to be a bug with mdast-usage? :/

chcokr:
Do you think it's a separate issue outside this PR?

wooorm:
Probably, I’ll investigate. mdast-usage is weird, because it in turn depends on `mdast` too.
I’m guessing it’s related to rerunning the cli multiple times though, so it’s uncovered with this feature, but it’s probably an error in my code!

chcokr:
OK, that must be complicated to debug :/ In the meantime, I'll just continue working on the spinner, while you investigate that bug!

wooorm:
Found it :) `mdast-usage` requires a file, and stores its `console.log`s, but the second time there are no `console.log`s of course!

Really looking forward to the spinned :tada:

chcokr:
Nice - I don't have to do anything in terms of that, right?

wooorm:
Nope!

chcokr:
Check out the spinner!

wooorm:
Very nice :dancer:

chcokr:
Do you know why the build is failing?

wooorm:
Hmm, it’s a minor issue that the spinner keeps moving down one line every save though...

Re failure: it’s expecting an empty string, but it gets part of a spinner... I’ll investigate

chcokr:
OK, I'll take care of the extra line while you investigate the build failure!

wooorm:
When without watch, one frame is rendered (I think, can you confirm?) and it’s never cleared.

I’d like the spinner to non-watch usage too, but I don’t think we should do it (I’d like the cli to be used programatically too, so other tools should be able to capture the output, which won’t work because the spinner litters the output). It’s probably best to output nothing when not watching!

chcokr:
> When without watch, one frame is rendered (I think, can you confirm?) and it’s never cleared.

I'll check.

> I’d like the spinner to non-watch usage too

I don't understand this sentence?

> It’s probably best to output nothing when not watching!

This is what we're already doing, right?

wooorm:
Re your last point: It would be awesome to have a spinner, especially when `mdast . -o` is taking a long time. But we shouldn’t do it. Everything is fine the way it is (except for that one frame of course!) :)

chcokr:
Ah OK. Yeah, I don't think showing a spinner in a non-watched `mdast` is a good idea.

Alright, let's see how long it takes to catch that extra frame...

wooorm:
I’m thinking there’s still a newline, and that that’s why the `quiet` fixture is complaining. Probably is the exact same thing that creates extra newlines on each `cmd+s`

chcokr:
Yeah, I'm trying to catch that.

chcokr:
I think I fixed it, can you check?

wooorm:
P.S. can you rebase with master? :)

chcokr:
Rebase done. 

Once we wrap up the spinner, the last thing we need to implement is this:

> Could we make sure not every file is reprocessed? It’s “nice to have” and not essential, but it would be very cool. 

This would be a great ending to our epic PR, but I have a question as to implementing it. How do I run *just one file* through the pipeline? I'm guessing I have to use `filePipeline` instead of `fileSetPipeline` - can you give me a code sample?

chcokr:
I need to leave, I'll be back in 3-4 hours.

wooorm:
Cool, I’m turning in for the night soon. Hope to fix this darn bash-test issue, and I’ll share my though on how to process just one file soon!

wooorm:
I just posted two PRs on your `watch` branch, they should get things working :)

When you’re done with that, can you squash your commits into one?

I’d like to run over all the changes one more time tomorrow, refactor some styles, &c.
When I’m done with that, feel free to also squash that commit with your changes.

And when that’s done, I’m in for merging this. We can add the “single changed file” processing later :+1:

chcokr:
PRs to a PR, that's the dream :D

Let me work on the things you mentioned, thanks.

chcokr:
Please see my comments at chcokr#1. I'm not squashing until we think more thru the `log-update` problem, in collaboration with sindresorhus/log-update#2 . Looks like he's bringing up a fix. Let me test his fix

chcokr:
`npm test` still fails when I use sindresorhus/log-update@a12769fdf663b59067a25d224be143c4df2f2005  as the dependency. @wooorm @sindresorhus

chcokr:
I'm going to bed, see you in about 10 hours

chcokr:
I have just now spoken with @wooorm. He tells me that he and @sindresorhus have spoken in person and are in the middle of discussing and bringing up a fix.

wooorm:
@chcokr Whoop, whoop! log-update@1.0.1 fixes this!

wooorm:
@chcokr I’m currently in the process of PRing a refactor (there are quite some code-style changes), and some wording things. I’m fine with a squash after that, and then, I think we can merge this!

chcokr:
Woohoo! I'll wait for your PR.

chcokr:
@wooorm What's the ETA for your PR?

wooorm:
@chcokr Yeah sorry, probably somewhere tomorrow!

chcokr:
:+1:

lib/cli/file-set.js
```diff
@@ -68,8 +68,20 @@ function one(fileSet) {
...
+     * @member {Array.<VFile>}
+     */
     self.contents = [];
+
+    /**
+     * Track the file-paths of files when they were added.
+     *
+     * @member {Array.<string>}
+     */
     self.sourcePaths = [];
```
chcokr: @wooorm Where in the codebase do `contents` and `sourcePaths` get used?

chcokr: Ohhh, nevermind, you were just adding JSDocs. Well, while I appreciate these additions, they are not relevant to the watch PR. Can we move them to a separate commit?

wooorm: We certainly can, I extracted them from your changes, which basically explained some confusing parts. Instead of having those descriptions all over the place, I instead opted for adding (rewritten) commits to the place where they are added.

I’m fine with both cases.

chcokr:
Everything looks good to me now. Should I squash?

wooorm:
Yup! Cool!

chcokr:
Man, 90 comments and 44 commits after! Woohoo

chcokr:
Time for merge :dancer:

lib/cli/watch-output-cache.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/**
+ * Add a VFile to the cache so that its content can later be used.
+ *
+ * @this {Cache}
+ * @param {VFile} file - The file to add to the cache.
+ */
+function add(file) {
+    var filePath = file.sourcePath || file.filePath();
```
chcokr: Oh shit hold on I forgot to get rid of `file.sourcePath ||`

wooorm:
Don’t worry, I’ll do some misc. refactoring anyway, so I can remove that too right?

wooorm:
And I’m just pushing the rebase now anyway :p

chcokr:
haha, alright :P

wooorm:
Nice work :D

PR #67, opened 2015-09-21
demo: interactive tree output

eush77:
This patch includes first attempts at interactive tree output for [mdast demo](http://mdast.js.org/demo/) page. This view is optional and not shown by default. Inspired by [fkling/esprima_ast_explorer](https://github.com/fkling/esprima_ast_explorer) and is sort of a stripped-down version of it.

Built with [jstree](https://github.com/vakata/jstree) which unfortunately requires jQuery to work. I'm happy to consider alternatives if this is too much overhead (`demo/build.js` increases in size from `247k` to `716k`, same ratio if minified).

Features:

- Each node in a tree is represented by its type (except for text nodes).
- Additional node attributes (such as `depth` on headings) are represented by key-value pairs next to node's type.
- A text node is represented by its value in quotes.
- Nodes can be opened and closed. Child AST nodes are represented by child nodes in a tree. Root node is opened by default.
- Source lexemes in the editor are selected on hover (this is rather hacky).

### DIFF
--- a/build.css
+++ b/build.css
@@ -1,0 +1,1471 @@
+/* jsTree default theme */
+.jstree-node,
+.jstree-children,
+.jstree-container-ul {
+  display: block;
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
+  list-style-image: none;
+}
+.jstree-node {
+  white-space: nowrap;
+}
+.jstree-anchor {
+  display: inline-block;
+  color: #333;
+  white-space: nowrap;
+  padding: 0 4px 0 1px;
+  margin: 0;
+  vertical-align: top;
+}
+.jstree-anchor:focus {
+  outline: 0;
+}
+.jstree-anchor,
+.jstree-anchor:link,
+.jstree-anchor:visited,
+.jstree-anchor:hover,
+.jstree-anchor:active {
+  text-decoration: none;
+  color: inherit;
+}
+.jstree-icon {
+  display: inline-block;
+  text-decoration: none;
+  margin: 0;
+  padding: 0;
+  vertical-align: top;
+  text-align: center;
+}
+.jstree-icon:empty {
+  display: inline-block;
+  text-decoration: none;
+  margin: 0;
+  padding: 0;
+  vertical-align: top;
+  text-align: center;
+}
+.jstree-ocl {
+  cursor: pointer;
+}
+.jstree-leaf > .jstree-ocl {
+  cursor: default;
+}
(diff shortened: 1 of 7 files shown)


wooorm:
Wow, very nice work! Thanks a lot!

I amended a little (I didn’t know I could add changes to someone’s commit!) to add an `assign` polyfill, and fixed a bug where the interactive tree wasn’t show (I changed from `if (!options.ast || !options.astTree) {` > `if (!options.astTree) {`).

Now, there’s still a bug where the image don’t get loaded (because they’re in `components/`, which is in `.gitignore`). Also: I just checked the homepage and the new styles seem to affect it too. I’ll figure this out

eush77:
Thank you!

> add an `assign` polyfill

I believe `Object.assign` polyfill is not necessary since [one](https://github.com/zloirock/core-js/blob/c841f4a251dd1aecfe04a6557cc5cd7c675db465/modules/%24.assign.js) is already [included](https://github.com/babel/babel/blob/75cd1a5531adb938f2747e4e5bb0928466246b1f/packages/babel/package.json#L49) in babel.

> I changed from `if (!options.ast || !options.astTree) {` > `if (!options.astTree) {)`

The idea behind that was that you check `Show AST` checkbox, and after that specify whether you need raw JSON or the tree with the next checkbox. I can't say it was intuitive, but neither is the current interface. Maybe change it to a radiogroup or tabs as in esprima_ast_explorer?

wooorm:
@eush77 For some reason it wasn’t included (for me), so Safari failed :/

Your last point: Maybe a select? `[yes] Output?`, with the options `Yes`, `JSON`, and `Tree`? Then we can disable all `stringify` settings (just like what’s now done for `show ast`) hen that setting isn’t `yes`.

eush77:
@wooorm I like it, but choosing between `Yes`, `JSON`, and `Tree` seems a bit weird don't you think? For `Tree` output, not only stringify settings don't make sense, but also positional information since it's needed for `mdast-range`.

wooorm:
Oh you’re right on `position:true` being required by `tree` :)

Do you have other suggestions for other terms, instead of `yes`, `json`, and `tree`?

eush77:
Maybe `processed`, `compiled`, or `output` instead of `yes`?

wooorm:
Or `markdown`?

eush77:
`markdown` is on point!

wooorm:
I just pushed a refactor on the output button (a lot better IMO)!

PR #69, opened 2015-09-22
demo: fix label on settings toggle

eush77:
Previously "Show settings" and "Hide settings" labels were mixed up.


wooorm:
Thanks! 😄

PR #70, opened 2015-09-23
demo: escape HTML text in tree nodes

eush77:
This patch escapes HTML special characters in a tree node's text content. Currently HTML nodes are broken and in general things can get ugly.


wooorm:
Thanks! I also committed some changes to the CSS to ensure the webfont rendered properly!

PR #72, opened 2015-09-23
demo: update mdast

eush77:
The demo is powered by the ancient `mdast@0.26`. This patch fixes it.


wooorm:
@eush77 Thanks! Totally forgot about that... :frowning:

PR #74, opened 2015-09-25
demo: save settings in localStorage

eush77:
Each time I use the demo page I have to set up the settings again (basically JSON or tree output), so why not save settings in localStorage? It is harmless, and I hope it would be useful to some.

### DIFF
--- a/build/demo.js
+++ b/build/demo.js
@@ -248,39 +248,94 @@ var debouncedChange = debounce(onchange, 20);
  * Setting changes.
  */
 
-function ontextchange($target, name) {
-    options[name] = $target.value;
+function targetType($target) {
+    var type = $target.hasAttribute('type') ? $target.type : $target.nodeName.toLowerCase();
+    return type in getTargetValue ? type : 'text';
 }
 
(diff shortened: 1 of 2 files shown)


wooorm:
Thanks! I opted for using a [simpler loop](https://github.com/wooorm/mdast/commit/cb62999ea88ca6ef1f7cf650d9b018c697eb891d#diff-96002143770e23013559d670d3b53871R199) instead; for some reason including Symbols doesn’t work with duo-babel

eush77:
It seems that the polyfill included in Babel is not [activated](http://babeljs.io/docs/usage/polyfill/), and that's why not every ES6 feature is available. I don't know if it is possible with `duo-babel` though, or even if it is a preferred workflow.

wooorm:
@eush77 Yup, for now I don’t mind just using the compilable sugar from esnext and omitting fancy polyfillable things. We could, when needed, [`require("babel/polyfill");`](http://babeljs.io/docs/usage/polyfill/#usage-in-node-browserify).

PR #78, opened 2015-10-01
Remove escape nodes

### LINKED ISSUE #71: What is the purpose of escape nodes?
I am wondering why escape nodes are in the AST? Are they ever useful?

Wouldn't it be simpler for users if

```
_foo\_ bar_
```

was parsed as

```json
{                                                                                   
  "type": "paragraph",                                                              
  "children": [                                                                     
    {                                                                               
      "type": "emphasis",                                                           
      "children": [                                                                 
        {                                                                           
          "type": "text",                                                           
          "value": "foo_ bar"                                                       
        }                                                                           
      ]                                                                             
    }                                                                               
  ]                                                                                 
}
```

and not as

```json
{
  "type": "paragraph",
  "children": [
    {
      "type": "emphasis",
      "children": [
        {
          "type": "text",
          "value": "foo"
        },
        {
          "type": "escape",
          "value": "_"
        },
        {
          "type": "text",
          "value": " bar"
        }
      ]
    }
  ]
}
```

It seems to me that escaping information is aimed primarily and solely at parser and plugins shouldn't even care. Commonmark spec [says](http://spec.commonmark.org/0.22/#backslash-escapes):

> Escaped characters are treated as regular characters and do not have their usual Markdown meanings.

In my interpretation, this means that both `_foo\_bar_baz_`, `_foo\_bar\_baz_`, and `_foo_bar_baz_` should result in the exact same syntax tree since in both cases inner underscores have no associated semantic meaning.

There is one use case for escape nodes that is special, I found, and that is using escape nodes with `value="\n"` as [hard line breaks](http://spec.commonmark.org/0.22/#hard-line-breaks). [Example](https://github.com/wooorm/mdast-html/blob/4315c6e95cf64027890ea4da79ec71ab76071eca/lib/compilers.js#L897-L899) from [mdast-html](https://github.com/wooorm/mdast-html):

```js
function escape(node) {
    return this[node.value === '\n' ? 'break' : 'text'](node);
}
```

But this is strange because mdast already has a break node. For example,

```
foo  
bar
```

(with two spaces at end of the first line) renders into

```json
{
  "type": "paragraph",
  "children": [
    {
      "type": "text",
      "value": "foo"
    },
    {
      "type": "break"
    },
    {
      "type": "text",
      "value": "bar"
    }
  ]
}
```

However,

```
foo\
bar
```

renders into

```json
{
  "type": "paragraph",
  "children": [
    {
      "type": "text",
      "value": "foo"
    },
    {
      "type": "escape",
      "value": "\n"
    },
    {
      "type": "text",
      "value": "bar"
    }
  ]
}
```

[According to](http://spec.commonmark.org/0.22/#hard-line-breaks) Commonmark both cases should be equivalent:

> For a more visible alternative, a backslash before the line ending may be used instead of two spaces.

So why not create break nodes in both cases and forego escape nodes?

My perspective on this is limited and maybe I'm getting it all wrong. Could you correct me then and point at the right direction?

eush77:
This patch changes the parser so that it constructs regular `text` nodes instead of `escape` nodes and changes the compiler so that it escapes characters all by itself.

As a result, AST is simplified (no more escape nodes) and compilation becomes more reliable (a plugin can safely insert asterisks into the tree and it won't affect the compiled Markdown in an unexpected way).

However, this is still a work in progress.

I decided to implement the absolutely simplest approach first — that is, to make stringifier escape _every_ escapable character. As @wooorm warned in https://github.com/wooorm/mdast/issues/71#issuecomment-143448370, this results in _a lot_ of backslashes in the output (for example, `npm run build-md` with this patch changes a lot of docs). Every single period is escaped. This issue is still to be addressed.

I open this PR for suggestions and feedback. Sorry it took me so long.

Close #71.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -33,6 +33,10 @@ var raise = utilities.raise;
 var clean = utilities.clean;
 var validate = utilities.validate;
 var normalize = utilities.normalizeIdentifier;
+var stateToggler = utilities.stateToggler;
+var noopToggler = utilities.noopToggler;
+var mergeable = utilities.mergeable;
+var MERGEABLE_NODES = utilities.MERGEABLE_NODES;
 var arrayPush = [].push;
 
 /*
@@ -76,7 +80,6 @@ var BLOCKQUOTE = 'blockquote';
 var LINK = 'link';
 var IMAGE = 'image';
 var FOOTNOTE = 'footnote';
-var ESCAPE = 'escape';
 var STRONG = 'strong';
 var EMPHASIS = 'emphasis';
 var DELETE = 'delete';
@@ -395,149 +398,6 @@ function getAlignment(cells) {
 
     return results;
 }
-
-/**
- * Construct a state `toggler`: a function which inverses
- * `property` in context based on its current value.
- * The by `toggler` returned function restores that value.
- *
- * @example
- *   var context = {};
- *   var key = 'foo';
- *   var val = true;
- *   context[key] = val;
- *   context.enter = stateToggler(key, val);
- *   context[key]; // true
- *   var exit = context.enter();
- *   context[key]; // false
- *   var nested = context.enter();
- *   context[key]; // false
- *   nested();
- *   context[key]; // false
- *   exit();
- *   context[key]; // true
- *
- * @param {string} key - Property to toggle.
- * @param {boolean} state - It's default state.
- * @return {function(): function()} - Enter.
- */
-function stateToggler(key, state) {
-    /**
-     * Construct a toggler for the bound `key`.
-     *
-     * @return {Function} - Exit state.
-     */
-    function enter() {
-        var self = this;
-        var current = self[key];
-
-        self[key] = !state;
-
-        /**
-         * State canceler, cancels the state, if allowed.
-         */
-        function exit() {
-            self[key] = current;
-        }
-
-        return exit;
-    }
-
-    return enter;
-}
-
-/**
- * Construct a state toggler which doesn't toggle.
- *
- * @example
- *   var context = {};
- *   var key = 'foo';
- *   var val = true;
- *   context[key] = val;
- *   context.enter = noopToggler();
- *   context[key]; // true
- *   var exit = context.enter();
- *   context[key]; // true
- *   exit();
- *   context[key]; // true
- *
- * @return {function(): function()} - Enter.
- */
-function noopToggler() {
-    /**
-     * No-operation.
-     */
-    function exit() {}
-
-    /**
-     * @return {Function}
-     */
-    function enter() {
-        return exit;
-    }
-
-    return enter;
-}
-
-/*
- * Define nodes of a type which can be merged.
- */
-
-var MERGEABLE_NODES = {};
-
-/**
- * Merge two text nodes: `node` into `prev`.
- *
- * @param {Object} prev - Preceding sibling.
- * @param {Object} node - Following sibling.
- * @return {Object} - `prev`.
- */
-MERGEABLE_NODES.text = function (prev, node) {
-    prev.value += node.value;
-
-    return prev;
-};
-
-/**
- * Merge two blockquotes: `node` into `prev`, unless in
- * CommonMark mode.
- *
- * @param {Object} prev - Preceding sibling.
- * @param {Object} node - Following sibling.
- * @return {Object} - `prev`, or `node` in CommonMark mode.
- */
-MERGEABLE_NODES.blockquote = function (prev, node) {
-    if (this.options.commonmark) {
-        return node;
-    }
-
-    prev.children = prev.children.concat(node.children);
-
-    return prev;
-};
-
-/**
- * Merge two lists: `node` into `prev`. Knows, about
- * which bullets were used.
- *
- * @param {Object} prev - Preceding sibling.
- * @param {Object} node - Following sibling.
- * @return {Object} - `prev`, or `node` when the lists are
- *   of different types (a different bullet is used).
- */
-MERGEABLE_NODES.list = function (prev, node) {
-    if (
-        !this.currentBullet ||
-        this.currentBullet !== this.previousBullet ||
-        this.currentBullet.length !== 1
-    ) {
-        return node;
-    }
-
-    prev.children = prev.children.concat(node.children);
-
-    return prev;
-};
 
 /**
  * Tokenise a line.  Unsets `currentBullet` and
@@ -1577,16 +1437,25 @@ function renderBlock(type, value, position) {
 /**
  * Tokenise an escape sequence.
  *
+ * Note that `$1` can only contain escapable characters per
+ * the used method, e.g., `\n` is only included when
+ * `commonmark: true`.
+ *
  * @example
  *   tokenizeEscape(eat, '\\a', 'a');
  *
  * @param {function(string)} eat
  * @param {string} $0 - Whole escape.
  * @param {string} $1 - Escaped character.
- * @return {Node} - `escape` node.
+ * @return {Node} - `text` node, or `break` node if
+ *   commonmark and the value is `'\n'`.
  */
 function tokenizeEscape(eat, $0, $1) {
-    return eat($0)(this.renderRaw(ESCAPE, $1));
+    if ($1 == '\n') {
+        return tokenizeBreak.call(this, eat, $0);
+    }
+
+    return eat($0)(this.renderRaw(TEXT, $1));
 }
 
 /**
@@ -2191,7 +2060,7 @@ function tokenizeFactory(type) {
         }
 
         /**
-         * Get offset. Called before the fisrt character is
+         * Get offset. Called before the first character is
          * eaten to retrieve the range's offsets.
          *
          * @return {Function} - `done`, to be called when
@@ -2342,7 +2211,7 @@ function tokenizeFactory(type) {
                     indent = combined.concat(indent);
                 }
 
-                node.position.indent = indent;
+                node.position.indent = indent || [];
 
                 return node;
             }
@@ -2386,7 +2255,9 @@ function tokenizeFactory(type) {
                 if (
                     prev &&
                     node.type === prev.type &&
-                    node.type in MERGEABLE_NODES
+                    node.type in MERGEABLE_NODES &&
+                    mergeable(prev) &&
+                    mergeable(node)
                 ) {
                     node = MERGEABLE_NODES[node.type].call(
                         self, prev, node
@@ -2429,10 +2300,12 @@ function tokenizeFactory(type) {
              * Add the given arguments, add `position` to
              * the returned node, and return the node.
              *
+             * @param {Object} node - Node to add.
+             * @param {Object} [parent] - Node to insert into.
              * @return {Node}
              */
-            function apply() {
-                return pos(add.apply(null, arguments), indent);
+            function apply(node, parent) {
+                return pos(add(pos(node), parent), indent);
             }
 
             /**
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -31,6 +31,9 @@ var defaultOptions = require('./defaults.js').stringify;
 
 var raise = utilities.raise;
 var validate = utilities.validate;
+var stateToggler = utilities.stateToggler;
+var mergeable = utilities.mergeable;
+var MERGEABLE_NODES = utilities.MERGEABLE_NODES;
 
 /*
  * Constants.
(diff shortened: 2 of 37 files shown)


lib/stringify.js
```diff
@@ -816,27 +881,19 @@ compilerPrototype.heading = function (node) {
...
- * @return {string} - Markdown escape.
- */
-compilerPrototype.escape = function (node) {
-    return '\\' + node.value;
+    if (
+        value[value.length - 1] == '\\' && parent &&
+        parent.type == 'linkReference' &&
+        parent.referenceType == 'shortcut'
+    ) {
+        // Don't escape trailing backslash in shortcut links.
```
eush77: This is here for [links-reference-style](https://github.com/wooorm/mdast/blob/b28a45ac3d8291ef2fb3e173468ac992f8c56f1f/test/tree/links-reference-style.json#L1702-L1734) test, but I suspect the test is wrong (there shouldn't be a `linkReference` node).

wooorm: I think you’re right, commonmark [disagrees](http://spec.commonmark.org/dingus/?text=%5C%5Ba%5D%20and%20%5Bb%5C%5D%0A%0A%5Ba%5D%3A%20http%3A%2F%2Fgoogle.com%0A%5Bb%5D%3A%20http%3A%2F%2Fgoogle.com) with mdast, and so does GH!

lib/parse.js
```diff
@@ -493,9 +492,31 @@ var MERGEABLE_NODES = {};
...
+            return true;
+        }
+
+        var start = node.position.start;
+        var end = node.position.end;
+
+        // Nodes with such a position info cannot be merged.
+        return start.line == end.line &&
+            end.column - start.column > node.value.length;
+    }
```
wooorm: Right, here you’re checking if these nodes represent a larger space than the actual size of their content right? Nice work!

Could you hoist it into the top-most scope? And, place it before their first use (I prefer my declarations before the spots they’re used :wink:)

test/index.js
```diff
@@ -650,7 +651,7 @@ describe('mdast.process(value, options, done)', function () {
     it('should run plugins', function () {
         assert(
             mdast.use(mentions).process('@mention') ===
-            '[@mention](https://github.com/blog/821)\n'
+            '[@mention](https://github\\.com/blog/821)\n'
```
wooorm: This seems to break, at least in [commonmark](http://spec.commonmark.org/dingus/?text=%5B%40mention%5D(https%3A%2F%2Fgithub%5C%5C.com%2Fblog%2F821)), the URL. Maybe we need to differentiate between plain-text escapes from links, language-flags, reference/definition identifiers?

wooorm:
Looks good, nice work; it seems like you understand the code really well already, I can imagine that that took some time, thanks for that! :smile:

I saw some places where you changed the `.`s in the input fixtures to comma’s, I understand why you did that, but I think that doing so makes it harder to see what changed without these escape nodes.

I think we should try to output as few slashes as possible, meaning, the output should look like the input in most of the current tests. For example, say the input is:

```md
foo-bar

foo
\-bar

**foo**
\-bar

~~-foo~~
```

These can be stringified correctly (without slash), by checking the following rules (`curr` is node with a possible escape character in it, rest is from that point):

If the character at position `0` in _curr_ is a list item bullet:

*   If _prev_ is a text node, and it doesn’t end in a newline, omit the slash;
*   If _parent_ is an inline node, omit the slash;
*   add the slash.

If the character at another position in _curr_ is a list item bullet:

*   If `charAt(pos-1)` is not a newline, omit the slash;
*   add the slash.

Now, there are of course more possibilities, such as the asterisk list-item bullet being interpreted as emphasis...

P.S. If we can get most characters to not be escaped, that would already be really good, e.g., the dash and the dot,
I don’t mind if some more obscure characters are always escaped.

test/index.js
eush77: @wooorm Your link includes two backslashes instead of one (two backslashes are needed to put it into a JS string, but the text includes only one). Commonmark actually [handles](http://spec.commonmark.org/dingus/?text=[%40mention]%28https%3A%2F%2Fgithub\\.com%2Fblog%2F821%29) escapes in links. It does not interpret escapes [in autolinks](http://spec.commonmark.org/dingus/?text=%3Chttps%3A%2F%2Fgithub\\.com%2Fblog%2F821%3E) but I think I handled that, and there're some tests for that.

wooorm: Oh my mistake, I remember it was either really early or really late when I posted that haha, sorry!

wooorm:
@eush77 I completely missed your latest commits, sorry! How’s it going with this? It looks like a lot of escapes are now “fixed”, as in, not showing up when not necessary. Are there still other escape characters you’re working on?

eush77:
I am currently in the middle of getting the entire test suite pass.

If you have some time, I would appreciate if you check out [this file](https://github.com/eush77/mdast/blob/escape-to-text/test/input/stringify-escape.text) and tell me what's missing there.

I also struggle with identifiers and shortcut references. For example, this input:

```
[two * three]
```

[parses into][example1]:

```
{
  "type": "paragraph",
  "children": [
    {
      "type": "linkReference",
      "identifier": "two * three",
      "referenceType": "shortcut",
      "children": [
        {
          "type": "text",
          "value": "two * three"
        }
      ]
    }
  ]
}
```

Normally asterisks in text nodes are escaped, but even if it wasn't the case inside `linkReference` nodes, this input:

```
[two \* three]
```

[leads][example2] to exactly the same tree (if `positions=false`; and it isn't the case in the demo, but it will), the only thing that differs is `identifier`.

Should I parse identifiers to decide what characters should get escaped and what shouldn't? Or is it ok that identifier will change after stringifying AST and parsing it again? I haven't found a solution yet.

[example1]: http://mdast.js.org/demo/?text=[two%20*%20three]
[example2]: http://mdast.js.org/demo/?text=[two%20\\*%20three]

wooorm:
I’ll take a look right now. Most important is that, but you maybe have figured this out already, a linkReference should be treated as either a link (when with definition), or as some sort of markdown!

wooorm:
Okay, to ensure the identifier is matched to a definition correctly, the `identifier` property must have the same treatment as a `definition`s `identifier` property. For example, see this:

```md
[two \* three]

[two * three]

[two * three]: http://google.com

[two \* three]: http://google.com
```

[Yields](http://mdast.js.org/demo/?text=%5Btwo%20%5C*%20three%5D%0A%0A%5Btwo%20*%20three%5D%0A%0A%5Btwo%20*%20three%5D%3A%20http%3A%2F%2Fgoogle.com%0A%0A%5Btwo%20%5C*%20three%5D%3A%20http%3A%2F%2Fgoogle.com):

```json
{
  "type": "root",
  "children": [
    {
      "type": "paragraph",
      "children": [
        {
          "type": "linkReference",
          "identifier": "two \\* three",
          "referenceType": "shortcut",
          "children": [
            {
              "type": "text",
              "value": "two "
            },
            {
              "type": "escape",
              "value": "*"
            },
            {
              "type": "text",
              "value": " three"
            }
          ]
        }
      ]
    },
    {
      "type": "paragraph",
      "children": [
        {
          "type": "linkReference",
          "identifier": "two * three",
          "referenceType": "shortcut",
          "children": [
            {
              "type": "text",
              "value": "two * three"
            }
          ]
        }
      ]
    },
    {
      "type": "definition",
      "identifier": "two * three",
      "title": null,
      "link": "http://google.com"
    },
    {
      "type": "definition",
      "identifier": "two \\* three",
      "title": null,
      "link": "http://google.com"
    }
  ]
}
```

We could of course apply the same escape-stripping system to both types of identifier (reference and definition) when compiling, for example, to:

```md
[two * three]

[two * three]

[two * three]: http://google.com

[two * three]: http://google.com
```

Or, we could treat both definitions as equal upon parsing (stripping the escapes in the identifiers).

Of course, this becomes interesting when the definitions point to different destinations!

wooorm:
P.S. I don‘t see a mention regarding escapes in [CommonMark’s matching](http://spec.commonmark.org/0.22/#matches) algorithm!

eush77:
Their dingus [matches](http://spec.commonmark.org/dingus/?text=[two%20\\*%20three]%0A%0A[two%20*%20three]%0A%0A[two%20*%20three]%3A%20http%3A%2F%2Fgoogle.com%0A%0A[two%20\\*%20three]%3A%20http%3A%2F%2Fyahoo.com) links with such identifiers correctly so I think we cannot afford stripping escapes. The most accurate solution is probably to parse identifiers alongside stringifying link's contents, but I'd like not to overcomplicate things if there's a better way.

wooorm:
@eush77 Ah right, well, that makes things a bit harder :anguished:

Maybe we can get away without this initially though, the cases where this would mess up horribly probably don’t occur much!?

eush77:
@wooorm 
> Ah right, well, that makes things a bit harder

Actually, it turned out to be rather simple: https://github.com/eush77/mdast/commit/baa1a13a9285cf6a44f0df7657a137942d88375d

I just rebased. Finally I think this is ready for review.

eush77:
Travis is green, but locally `npm run build-md` fails because `mdast-html` compiler [doesn't pass][1] a `node` argument to `this.encode`. This [could've led][2] to errors before, but with this patch `this.encode` [relies][3] on `node` much more heavily (and also on `parent`, which is something I had to add).

I will move escaping code out of `this.encode` to fix this, but long-term it could be better to extract the encoding function other compilers might need to a `mdast-util` module and make them depend on it instead of using `this.encode`, or at least make `this.encode` part of the API and document somewhere.

[1]: https://github.com/wooorm/mdast-html/blob/4315c6e95cf64027890ea4da79ec71ab76071eca/lib/compilers.js#L882
[2]: https://github.com/wooorm/mdast/blob/0ec8dede49e7c7f5433531408ddcd0e7d7ada772/lib/stringify.js#L276
[3]: https://github.com/eush77/mdast/commit/b62e541fa061745cafbab4f55e968450b0581f26#diff-dd5072864997b356b1727023f837c921R343

wooorm:
Looks very good @eush77, I’d like to look over the code a bit, maybe I’d be able to do that tonight. I’ll probably have some code-style related things, and I’ll submit a PR on your mdast fork. Then, we can squash the commits together to have one nice commit!

wooorm:
@eush77 Ping! Sorry, I totally forgot I was supposed to look over this. I will, but before I do, was there more stuff you were planning on including in this PR?

eush77:
@wooorm Sorry, it's my fault. I was supposed to [rewrite](https://github.com/eush77/mdast/pull/1) some parts using no (or less) regexps in order to get more predictable performance. And I did that… partially, but I also want to profile these parts in order to avoid the trap of premature optimization. I will try to finish it soon.

wooorm:
Oh right, totally forgot :smile: What’s the estimate ETA for your work? Somewhere the coming week?

eush77:
@wooorm Yeah, something like that :smile:

wooorm:
Great! I’m *almost* done with the entities part, together with the escapes we can release mdast@3.0.0!

eush77:
@wooorm Okay, I think this is ready for another review.

wooorm:
:+1: looks like great work!

wooorm:
I'll try to respond tomorrow!

wooorm:
Nice work. I just had a an idea, for example, with the tildes.

Would it be a good idea to escape double tildes in *any* escaped-tilde supporting environment? So, in commonmark mode too? Otherwise, when someone posts the markdown on GitHub, it would render as ~~strikethrough~~.

On the other hand, that would break support for “normal” markdown...

eush77:
> Would it be a good idea to escape double tildes in _any_ escaped-tilde supporting environment? So, in commonmark mode too?

Since GFM is on by default, it works in Commonmark too, unless you explicitly turn GFM off:

```js
> console.log(JSON.stringify(ast, null, 2))
{
  "type": "root",
  "children": [
    {
      "type": "paragraph",
      "children": [
        {
          "type": "text",
          "value": "~~foo~~"
        }
      ]
    }
  ]
}
> mdast.stringify(ast, { commonmark: true })
'\\~~foo\\~~\n'
> mdast.stringify(ast, { commonmark: true, gfm: false })
'~~foo~~\n'
```

Maybe we could escape double tildes in Commonmark even with `gfm: false`, but the use case is not clear to me. Why would anyone post markdown to GitHub and compile if with `gfm: false`? I'm not objecting to it but I'm not sold on its usefulness either.

wooorm:
Ah, the fact that someone explicitly sets it to false seems good enough for me to *not* do this.

Reason for doing that was “cross-vendor support”, but as this idea would break normal markdown (because of the extra slash), that wouldn’t really count.

Never mind! :)

wooorm:
Great work! I have some very small style change suggestions but I’ll do that later.

Most important: how is your overall feeling related to this? Does it work as expected?

Also: It’s quite hard to get the picture of what exactly changed, I think squashing everything might help me. Are you interesting in doing that already, or would you like to continue working some more!

wooorm:
@eush77 I’m probably one or two days away from merging this. Any last notes you’d like to add (see previous comment).

eush77:
@wooorm To my knowledge, the only thing that is still not escaped properly is pipes in GFM mode (which is also of course trivial to fix). Docs and tests for the remaining modes (currently it's just GFM and Commonmark+GFM) are missing. But mostly I think this is done. I will try to rebase shortly and probably deal with remaining tests later.

wooorm:
Cool! Thanks @eush77. I might be able to work on the rebase too. The parser has been completely rewritten so that will probably be a huge hassle, but, the compiler remained untouched.

I also added a list of [escapes as a JSON file](https://github.com/wooorm/mdast/blob/next/lib/escape.json), which might be useful?

wooorm:
@eush77 Well, that looks fixed! I’m quite eager to work on this. Ping me if you get around to squashing/rebasing :ok_hand:

eush77:
@wooorm I squashed and rebased this branch on `master` and I'm rebasing on `next` in a [separate branch](https://github.com/eush77/mdast/commits/escapes-next) — the only nontrivial conflicts seem to be with tests.

wooorm:
Cool, great work!

Maybe this helps: I find both `node script/regenerate-fixtures.js` and `npm run test-api-extensive` (which tries all available parsing options on every fixture) to be very useful. Do however make sure to check whether all changes ASTs are correct!

eush77:
@wooorm Thank you!

Good to know that `script/regenerate-fixtures.js` is built in. I have done this manually many times.

wooorm:
Yes, I found myself repeatedly doing that a lot So I built this in. That, and `npm run test-api-extensive` is very useful for detecting changes :+1:

wooorm:
Wow! :dancer:

Really great work @eush77! :kissing_heart: I did some refactoring, but not that much!

P.S. Ignore the travis failure, that’s just old plug-ins.

eush77:
:tada:

Issue #73, opened 2015-09-24
Tables containing code with pipe characters are formatted wrong


anko:
### Test case

`test.markdown`:
```markdown
| abc | head2 |
| ---| ----- |
| x     | `|||` |
```

Command:

```
mdast < test.markdown
```

### Expected output

```
| abc | head2 |
| --- | ----- |
| x   | `|||` |
```

### Observed output

```
| abc | head2 |     |
| --- | ----- | --- |
| x   | `     | `   |

 |


<stdin>
       3:13  error    Incorrectly eaten value: please report this warning on http://git.io/vUYWz
       3:14  error    Incorrectly eaten value: please report this warning on http://git.io/vUYWz

✖ 2 errors
```

### Supplementary info

```
$ mdast -V
2.0.0
```

wooorm:
Thanks @anko for reporting this!

I remember markdown-it having a similar problem. They came up with an approach which I implemented here (although, as the warnings suggest, not correctly…)
The errors should of course be fixed, just like the weird newline in the last cell. However, I’m not sure if it’s possible to
parse the pipes in inlineCode in tableCells the way you expected.
The current system first gathers the cells (before there’s knowledge of inlineCode). Later, the content of cells is parsed as inline.

I’ll look into it this weekend

wooorm:
Sorry that I didn’t get back to you earlier, hope I have some time this weekend :anguished:

wooorm:
Well, I fixed the incorrect eating errors, but I haven’t fixed the pipes in inline-code yet. The only viable system for that seems to be to actually pre-parse the content in each cell for inline-code :anguished:

PR #83, opened 2015-10-03
Fix multiple adjacent empty table-cells

wooorm:
This will replace the current frail regex-based cell parsing system.

However, this does not address the case where pipes occur in `inlineCode`

Related: GH-73.


wooorm:
Closes by c7e6fdb.

PR #85, opened 2015-10-05
Watch: process just the changed file

chcokr:


lib/cli/file-set-pipeline/traverse.js
```diff
@@ -27,6 +27,10 @@ var Traverser = require('../traverser');
  * @param {CLI} cli
  */
 function traverse(cli) {
+    if (cli.changedFilePath) {
+        return;
```
chcokr: @wooorm So here's what I'm talking about. Try the following.

Test case 1:
- Run `mdast -w readme.md`, with no errors in `readme.md`.
- Make an error in `readme.md`. No warnings show up. This is a bug.

Test case 2:
- Before you run mdast, make a line in `readme.md` longer than 80 characters. Make it like 90 characters.
- Run `mdast -w readme.md`, with the error still in `readme.md`.
- You get a warning upon mdast's load. This is correct.
- Delete one character at the end of that 90-character line. You'll notice the warnings from prior runs pile up. This is a bug.

chcokr: These bugs don't show up if you comment out L31.

wooorm:
It seems like the files in the file-set don’t change when running the CLI another time. I suggest creating a new file-set and adding just the changed files!

chcokr:
Ah OK. Let me get to it. Other than, how does this PR look?

chcokr:
Wait, actually, looking at `transform.js`, I think what you said is exactly what I'm doing in `transform.js` - am I wrong?

chcokr:
I'm going to bed now, looking forward to your comments.

wooorm:
:+1:

I think you’ll need to reset tome other properties too, e.g., `cli.files = []`, not sure yet about others!

chcokr:
Thanks for your PR, very elegant :smile: The change set is so small now, haha.

Something I didn't quite understand from your PR:

> Currently, changing mdastrc(5) files is ignored (the config has a cache by file-path at cli.config.cache). Changing mdastignore(5) files however do take affect, as they are reloaded on each run. I’m not sure whether any choice is good here: having no caching can be expensive, having to exit a watch for changes to take affect isn’t very user-friendly either.

Can you give me a step-by-step procedure to reproduce the problem? I just don't understand the problem.

chcokr:
I went ahead and squashed the commits.

lib/cli/file-set-pipeline/configure.js
```diff
@@ -28,16 +28,16 @@ var Configuration = require('../configuration');
...
-        'detectRC': cli.detectRC,
-        'file': cli.configPath,
-        'settings': cli.settings,
-        'plugins': cli.plugins,
-        'output': cli.output,
-        'cwd': cli.cwd
-    });
-
-    cli.configuration = configuration;
+    if (!cli.configuration) {
```
chcokr: @wooorm I'm actually not sure if the changes in this file are necessary. Everything seems to work just fine without these changes.

wooorm: Things will work fine (this is also related to your previous question), however, by wrapping in an if-statement no new 'congifuration' instance is created on subsequent 'run's.

You should be able to reproduce this by changing a lint rule value in .mdastrc (see mdast-lint for how to do that), E.g., setting maximum-line-length to 20. Theoretically, this change shouldn't start warning because the configuration is cached in memory and not reloaded! But, this isn't true if you change a pattern in the .mdastignore file as it isn't cached.

Did that help? :)

chcokr: Ah OK, thanks for the clarification. Now it makes sense.

Yeah, we should keep these changes.

chcokr:
Alright, I'd call it ready for merge!

lib/cli/file-set-pipeline/configure.js
wooorm: In fact, we can probably also wrap the traverser inside a similar if-statement (but keep the 'CLI.files' outside it of course), that should ensure mdastignore and mdastrc are treated similarly (as in, cached). Can you try that out?

chcokr: On it.

lib/cli/file-set-pipeline/traverse.js
```diff
@@ -27,14 +27,15 @@ var Traverser = require('../traverser');
...
-    cli.files = traverser.traverse(cli.globs);
+    if (!cli.files) {
+        cli.traverser = new Traverser({
+            'extensions': cli.extensions,
+            'ignorePath': cli.ignorePath,
+            'detectIgnore': cli.detectIgnore,
+            'cwd': cli.cwd
+        });
+    }
+    cli.files = cli.traverser.traverse(cli.globs);
```
chcokr: @wooorm Check this out. We have to cache the traverser, so I went with the `cli.traverser` approach.

chcokr: Now that I think about it, `if(!cli.traverser)` would be better..

wooorm: I agree, check CLI.traverser seems more logical!

chcokr:
@wooorm LGTY now?

wooorm:
Yup! Looks fine!

chcokr:
Thanks, I squashed the commits.

wooorm:
:+1:

chcokr:
This does make working with `mdast -w .` a lot more pleasant :smile:

wooorm:
Agreed @chcokr, very much so and very much faster!

Thanks for the work :)

PR #101, opened 2015-12-01
cli: don't check for plugins key recursively

### LINKED ISSUE #99: problems with check for PLUGIN_KEY in merge in lib/cli/configuration.js
so basically if mdast plugin also has `plugins=["foo", "bar"]` option, then [`merge` function][merge] and [check for `PLUGIN_KEY=plugins`][check] will turn them to `plugins={ foo: null, bar: null }`

*.mdastrc:*
```json
{
	"plugins": {
		"mdast-demo": {
			"plugins": [ "foo", "bar" ], 
			"plugins2": [ "foo", "bar" ]
		}
	}
}
```
*mdast-demo:*
```js
module.exports = function attacher(processor, options) {
  return function transformer(ast) {
    console.log('plugins', options.plugins);
    console.log('plugins2', options.plugins2);
  };
};
```

*log:*

    plugins { foo: null, bar: null }
    plugins2 [ 'foo', 'bar' ]

[merge]: https://github.com/wooorm/mdast/blob/master/lib/cli/configuration.js#L69
[check]: https://github.com/wooorm/mdast/blob/master/lib/cli/configuration.js#L81

eush77:
Previously, configuration merger would turn all "plugins" arrays into objects, even inside plugin option objects. This patch checks for this property only at the top scope and not recursively.

Fix #99.

### DIFF
--- a/lib/cli/configuration.js
+++ b/lib/cli/configuration.js
@@ -66,7 +66,7 @@ extend(base.settings, defaults.stringify);
  * @param {Object} configuration - Configuration to merge from.
  * @return {Object} - `target`.
  */
-function merge(target, configuration) {
+function merge(target, configuration, recursive) {
     var key;
     var value;
     var index;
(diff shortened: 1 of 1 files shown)


wooorm:
Thanks for the great patch, @eush77 :ok_hand:

iamstarkov:
yeah, @eush77 its cool, thank you

iamstarkov:
@wooorm would you mind to publish mdast with these changes?

PR #107, opened 2015-12-13
Add link to mdast-util-inject

anandthakker:
Inserted it in alpha order by the package name, since they already seemed to be ordered that way.


anandthakker:
Closing in favor of rebased version #108

PR #108, opened 2015-12-13
Add link to mdast-util-inject

anandthakker:
Inserted it in alpha order by the package name, since they already seemed to be ordered that way.


wooorm:
Thanks for the addition!

Please add an `em` dash (`—`) instead of a normal dash (which is seen as a list-item bullet).
Also, make sure no line is longer than 80 characters.

Most importantly, please don’t open many PRs. I can do rebasing / squashing by hand if need be, but I’d prefer not to have too many empty issues / PRs in the Git history :+1:

anandthakker:
:+1: switched to em dash, wrapped to 80 chars.

wooorm:
Thanks @anandthakker! :ok_hand:

PR #139, opened 2016-01-30
Improve parsing a URL in text

MakeNowJust:
Current implementation parsed following text

```markdown
http://foo.example.com
```

as such a following code.

```markdown
<http://foo.example>.com
```

Almost all people don't expect such a behavior, I fixed it.


wooorm:
Thanks, working on this!

wooorm:
Thanks! Great work :+1:

PR #150, opened 2016-02-07
stringify: fix entities in shortcut and collapsed references

### LINKED ISSUE #141: Entities in references and definitions
Entities in references and definitions are currently compiled in a way which removes their bond:

```md
[foo&amp;bar][]

[foo&amp;bar]: http://example.com
```

Is compiled to:

```md
[foo&bar][]

[foo&amp;bar]: http://example.com
```

@eush77, I think this has to do with [`unescapeShortcutLinkReference`](https://github.com/wooorm/remark/blob/8ff7ab436cdcd0daf0f1210fb65ad46976449bdc/lib/stringify.js#L1828), which you you added in d1fe019. Could you take a look at this?

eush77:
This patch fixes #141.

Previously `unescapeShortcutLinkReference` was called only for shortcut references (hence the name), and it was reverting some of escaping that was done before. This patch removes the unnecessary work and expands `unescapeShortcutLinkReference` to HTML entities as well as collapsed references.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -287,28 +287,44 @@ function encodeFactory(type) {
 }
 
 /**
- * Check if a string starts with HTML entity.
+ * Returns the length of HTML entity that is a prefix of
+ * the given string (excluding the ampersand), 0 if it
+ * does not start with an entity.
  *
  * @example
- *   startsWithEntity('&copycat') // true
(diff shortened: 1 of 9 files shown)


wooorm:
Looks good! Thanks @eush77!

Issue #126, opened 2016-01-07
cli: AST input


eush77:
Currently `remark(1)` can process in `markdown⟶markdown` mode or `markdown⟶AST` (with `-a --ast`).

It would be useful to support some kind of AST input so that `AST⟶markdown` and `AST⟶AST` operation modes would be possible, too.

This would be useful for quickly processing documents with JSON processors (like [jq]) when a plugin does not exist / is not general enough or for testing plugins on predefined or manually constructed mdast trees.

[jq]: https://stedolan.github.io/jq/

wooorm:
Definitely! This would additionally be a great way to patch remark/retext/hast together!  For example, `remark -a -u lint | to-nlcst | retext -a -u spell -s | to-mdast | remark -s`.

However, both `-s` and `-S` are already in use. And so is `-c` (for compile). What would be a great name?

eush77:
`-m --mdast` or `-r --raw` are both unused.

But then there is an issue with `-a --ast` (I think): when `remark(1)` gets the feature of reading AST input, `--ast` becomes ambiguous. So the options for AST input/output should probably come in pair: something like `--ai --ast-input` / `--ao --ast-output`. `-a --ast` could enable both.

wooorm:
Oh right. So `-a` should stand for both `--ast-in` and `--ast-out`, agreed.
We could opt for no short flags for these two.

Maybe though `-t` would be better (as the same could be used for retext, which has a CST): `-t`, `--tree-in`, and `--tree-out`?
That would result in `remark readme.md -u lint --tree-out | to-nlcst | retext -u spell -t | to-mdast | remark --tree-in`.

eush77:
Both `--ast` and `--tree` are fine to me. I can also suggest `--json`.

PR #156, opened 2016-02-10
Add support for syntax tree input to remark(1)

wooorm:
This update removes the previously supported `-a, --ast` flag, and
renames it to `--tree-out`.  In addition, `--tree-in` is now supported
to depict that the input read from stdin(4) is an already parsed
syntax tree in JSON.  And, this commit introduces a new `-t, --tree`
flag which turns both `--tree-in` and `--tree-out` on, thus specifying
a tree input on stdin(4) and a tree output on stdout(4).

Closes GH-126.
Closes GH-156.

### DIFF
--- a/doc/remark.1.md
+++ b/doc/remark.1.md
@@ -132,13 +132,30 @@ $ remark --no-rc readme.md -oqw
 The watch is stopped when `SIGINT` is received (usually done by pressing
 `CTRL-C`).
 
-### `-a`, `--ast`
+### `-t`, `--tree`
 
 ```sh
-remark readme.md --ast
+remark --tree < input.json > output.json
 ```
 
-Instead of outputting document the internally used AST is exposed.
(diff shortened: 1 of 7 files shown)


wooorm:
@eush77 Can you review whether this closes your issue?

wooorm:
Just saw that the docs show the same example, that should at least be fixed.

eush77:
Yeah, this is cool! :smile: 

One thing, somewhat unexpectedly:

```fish
$ remark --tree-in b.json
Error: remark does not accept both files and `--tree-in`
```

Why is that?

wooorm:
For some weird reason I initially only allowed stdin(4) as a syntax tree, I changed that, so now it’s possible to transform files from the file-system too: `remark --tree fixtures/tree -e json -o` should in theory rewrite a directory of syntax tree fixtures.

PR #162, opened 2016-02-16
Do not escape emphasis markers in words

### LINKED ISSUE #161: Do not escape emphasis markers in words
Currently, emphasis markers inside words (`foo_bar`) are escaped (`foo\_bar`).

The reason for this is to enable markdown processed by remark to render correctly in older markdown processors (as in, those strictly adhering to Gruber’s original markdown), but it is starting to get in the way of more progressive flavours (as in, GFM), as can be seen in wooorm/remark-lint#46.

Therefore, I want to remove the escaping of emphasis markers inside words by default, and hide it behind a `pedantic` compilation setting.

eush77:
This patch removes excessive escaping of underscore characters when surrounded by alphanumerics (not needed unless in pedantic mode).

Close #161.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -110,6 +110,16 @@ var TICK = '`';
 var TILDE = '~';
 var UNDERSCORE = '_';
 
+/**
+ * Check whether `character` is alphanumeric.
+ *
+ * @param {string} character - Single character to check.
+ * @return {boolean} - Whether `character` is alphanumeric.
+ */
+function isAlphanumeric(character) {
+    return /\w/.test(character) && character !== UNDERSCORE;
+}
+
 /*
  * Entities.
  */
@@ -409,6 +419,7 @@ function escapeFactory(options) {
         var self = this;
         var gfm = options.gfm;
         var commonmark = options.commonmark;
+        var pedantic = options.pedantic;
         var siblings = parent && parent.children;
         var index = siblings && siblings.indexOf(node);
         var prev = siblings && siblings[index - 1];
@@ -419,6 +430,8 @@ function escapeFactory(options) {
         var escaped = queue;
         var afterNewLine;
         var character;
+        var wordCharBefore;
+        var wordCharAfter;
 
         if (prev) {
             afterNewLine = prev.type === 'text' && /\n\s*$/.test(prev.value);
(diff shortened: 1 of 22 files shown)


lib/stringify.js
```diff
@@ -110,6 +110,16 @@ var TICK = '`';
...
 
+/**
+ * Check whether `character` is alphanumeric.
+ *
+ * @param {string} character - Single character to check.
+ * @return {boolean} - Whether `character` is alphanumeric.
+ */
+function isAlphanumeric(character) {
+    return /\w/.test(character) && character !== UNDERSCORE;
+}
```
eush77: Should be `isAlphabetic(character) || isNumeric(character)` but that requires extracting these functions and characters from `parse.js` (out of scope of this PR).

test/index.js
```diff
@@ -1277,7 +1278,7 @@ describe('fixtures', function () {
 
                     compare(node, trees[mapping[key]], false, initialClean);
 
-                    markdown = remark.stringify(node, stringify);
+                    markdown = remark.stringify(clone(node), stringify);
```
wooorm: @eush77 Why was this needed?

lib/stringify.js
wooorm: :+1:

test/index.js
eush77: This is needed for some [stringify tests](https://github.com/wooorm/remark/blob/3802139667b336bdcbcfdc7f7859391456ab0d37/test/index.js#L1285) because node mergers currently [modify AST nodes](https://github.com/wooorm/remark/blob/3802139667b336bdcbcfdc7f7859391456ab0d37/lib/utilities.js#L257) during compilation, and because of that it can sometimes become invalid. I would suggest returning new nodes and avoid mutations in compiler instead. What do you think?

eush77: Or I guess mutations would be ok if we could just remove merged nodes from AST.

wooorm: Ah ok :ok_hand:

I’d opt for the latter, or urge users to do their own cloning if they really want to snapshot a tree. Especially for big documents, cloning so often could be quite heavy!

eush77:
Agreed :smile:

test/index.js
eush77: Agreed :smile:

wooorm:
Looks good. Merging. And, releases as 4.1.0.

undgrnd:
Hello,

how can I disable escaping of emphasis character (underscore) in remark-stringify?

I use stringify like this:

remark() .use({ settings: { listItemIndent: 1, }, }) .stringify({ type: 'root', children: newAst });

And I get "page\\_\\_title" instead of "page__title".

This option does not affect anything: https://github.com/remarkjs/remark/tree/master/packages/remark-stringify#optionspedantic

PR #167, opened 2016-02-24
Strip merged nodes from AST in compiler

eush77:
This patch removes AST nodes that are merged in compiler from the tree. Currently it is not the case, and so AST can be messed up during the process.

To illustrate the problem (see also https://github.com/wooorm/remark/pull/162#discussion_r52991739):

`ast`:

```json
{
  "type": "paragraph",
  "children": [
    {
      "type": "text",
      "value": "foo"
    },
    {
      "type": "text",
      "value": "bar"
    },
    {
      "type": "text",
      "value": "baz"
    }
  ]
}
```

```js
remark.stringify(ast);
```

`ast`:

```json
{
  "type": "paragraph",
  "children": [
    {
      "type": "text",
      "value": "foobarbaz"
    },
    {
      "type": "text",
      "value": "bar"
    },
    {
      "type": "text",
      "value": "baz"
    }
  ]
}
```

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -1007,6 +1007,7 @@ compilerPrototype.all = function (parent) {
     var values = [];
     var index = 0;
     var length = children.length;
+    var mergedLength = 1;
     var node = children[0];
     var next;
 
@@ -1027,10 +1028,12 @@ compilerPrototype.all = function (parent) {
         } else {
             values.push(self.visit(node, parent));
             node = next;
+            children[mergedLength++] = node;
         }
     }
 
     values.push(self.visit(node, parent));
+    children.length = mergedLength;
 
     return values;
 };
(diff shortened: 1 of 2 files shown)


wooorm:
Wait, is `foobarbaz`, `bar`, `baz` the real output?

Great otherwise :)

wooorm:
Are you saying that happens now? Hmm, darn.

eush77:
No, the output is fine (just `foobarbaz`), but `ast` contains these extra nodes after `remark.stringify` call.

Succeeding `remark.stringify` calls would return garbage (`foobarbazbarbaz`, then `foobarbazbarbazbarbaz` &c) :smile:

wooorm:
Oh right, I see, awesome!

wooorm:
Thanks a lot, @eush77, I won’t release it yet though, maybe there are other things coming up this week too!

PR #171, opened 2016-03-30
Enable using JS file as configuration file

### LINKED ISSUE #170: Configuration file as JavaScript file
How about allowing to use JavaScript as the configuration file, which would `module.exports = {};` the configuration?
In the same manner as ESLint (http://eslint.org/docs/user-guide/configuring#configuration-file-formats) or StyleLint (http://stylelint.io/user-guide/configuration/) does it...

Related to https://github.com/wooorm/remark/blob/master/doc/remarkrc.5.md

paazmaya:
Example of such configuration file included as the `.remarkrc-man.js`.

Fixes #170


wooorm:
Nice first work, but you’re missing support for detecting `.remarkrc.js` files. E.g., when executing `remark .`, `.remarkrc.js` files are not detected currently. I’m working on it.

paazmaya:
Thanks for releasing this!

PR #205, opened 2016-08-30
rename fixtures with * in path to use asterick

### LINKED ISSUE #204: Test fixtures with asterisks in the filename can't be cloned on Windows
The invalid characters are listed on [msdn](https://msdn.microsoft.com/en-au/library/windows/desktop/aa365247(v=vs.85).aspx#naming_conventions)

The `*` is the problem character

Proposed solution is to just use 'asterisk' and the like in filenames, and then translate to the charachters

BarryThePenguin:
fixes #204


wooorm:
@BarryThePenguin Oh thanks so much, I appreciate it!

So, does remark now build on Windows?

BarryThePenguin:
I can clone it on windows now. I haven't tried building it, the only script I've run is `test-coverage`. lerna doesn't bootstrap without admin privilege. That's about as far as I've got...

wooorm:
Oh really? Hmm. I use nvm/n, which are really useful and fix the admin things. Maybe with [nvm-windows](https://github.com/coreybutler/nvm-windows) it works?

But great the tests run now :)

BarryThePenguin:
Oh.. nah, even with an environment/permissions setup properly, you still need admin privileges to create symlinks :( lerna/lerna#319 and lerna/lerna#237 explain a bit

Issue #219, opened 2016-12-15
remark: command not found when using remark-cli


gotofritz:
Hi. I was trying to use stylelint-config-standard, which uses remark-cli, and I when I tried to run `npm test` I got `"remark: command not found"`.

I explored a bit and it looks like npm doesn't like the way the path for bin in the remark-cli's package.json is set (it just needs to start with "./"). I will submit a PR.

wooorm:
How weird, are you sure about this?
I’ve used this on Node 4 extensively, and never ran into any problems!

gotofritz:
Positive, it happens with 4.2.6 for me

wooorm:
Funky: I just cloned `stylelint-config-standard`, and did the following:

```sh
nvm use 4.2.6
cd stylelint-config-standard
npm i
```

...and everything went fine!

Could you show me how I can reproduce this problem?

gotofritz:
Mmm. A restart seems to have automagically fixed everything O_o

Apologies for wasting your time

wooorm:
No problem, thanks for checking!

sidvenu:
I am facing this issue currently. It says `remark: command not found` when I try running it.

wooorm:
@sidvenu Please open new issues and fill in the issue template.

PR #220, opened 2016-12-15
Changed path of bin in package.json of remark-cli to avoid errors in node 4

gotofritz:
See https://github.com/wooorm/remark/issues/219


wooorm:
Closed per GH-219.

PR #226, opened 2017-01-13
Change min row/column from 2 to 1

binhnd-socicom:


wooorm:
Please supply more info: Why should this be included, what does it do, new tests, changed fixtures, etc.

wooorm:
@binhnd-socicom Ping!

binhnd-socicom:
Oh sorry. I missed your messages.
The limitation causes table with single column cannot be displayed correctly. 
I forked your branch into my own to modify it directly and it's working as my expectation.

Thank you for your repo.

PR #229, opened 2017-01-30
Added paddedTable to remark-stringify

alex-e-leon:
@wooorm in reference to https://github.com/wooorm/markdown-table/pull/9#issuecomment-275703794

I'm not sure if `paddedTable` or `padTable` is better, but I just followed your convention from `looseTable` and `spacedTable`.

Heads up - this is a bigger project to wrap my head around, so I mostly just followed convention rather than trying to figure out what was actually going on. TBH I haven't even tested this against a real markdown file. That said, I think this works?


wooorm:
Sorry for the wait. I’ll release this in a minute!

alex-e-leon:
@wooorm No problem, thanks for your help! - Do you mind pinging me here when you've released a new version?

wooorm:
There’s a new version of remark-stringify out, so you can update to that!

PR #262, opened 2017-06-09
[WIP] es6

Rokt33r:
Sorry for my late response.

## Done 

- Use const and let rather than var
- Remove 'use strict'
- Use es6 module
- Enable jsnext
- Disable `prefer-const`
  Some of variables defined only once. So, xo is asking us to use const. If we do, the current convention, putting every variables on top, will be broken.

## To do

- Discard redundant module, like xtend.(Object.assign can do same thing)
- Introduce modern build(webpack/rollup?)/testing(Jest?)
- ... and more

I need more feedback before we decide how should we refactor this. 
Also, I'm considering to rewrite whole code as a typescript or to prepare solid type definition.

### DIFF
--- a/packages/remark-parse/lib/locate/escape.js
+++ b/packages/remark-parse/lib/locate/escape.js
@@ -6,9 +6,7 @@
  * @fileoverview Locate an escape.
  */
 
-'use strict';
-
-module.exports = locate;
+export default locate;
 
 function locate(value, fromIndex) {
   return value.indexOf('\\', fromIndex);
--- a/packages/remark-parse/lib/parser.js
+++ b/packages/remark-parse/lib/parser.js
@@ -5,17 +5,14 @@
  * @module remark:parse
  * @fileoverview Markdown parser.
  */
+import xtend from 'xtend';
+import toggle from 'state-toggle';
+import vfileLocation from 'vfile-location';
+import unescape from './unescape';
+import decode from './decode';
+import tokenizer from './tokenizer';
 
-'use strict';
-
-var xtend = require('xtend');
-var toggle = require('state-toggle');
-var vfileLocation = require('vfile-location');
-var unescape = require('./unescape');
-var decode = require('./decode');
-var tokenizer = require('./tokenizer');
-
-module.exports = Parser;
+export default Parser;
 
 /* Construct a new parser. */
 function Parser(doc, file) {
@@ -35,7 +32,7 @@ function Parser(doc, file) {
 }
 
 /* Prototype. */
-var proto = Parser.prototype;
+const proto = Parser.prototype;
 
 /* Expose core. */
 proto.setOptions = require('./set-options');
@@ -151,8 +148,8 @@ proto.tokenizeFactory = tokenizer;
 
 /* Get all keys in `value`. */
 function keys(value) {
-  var result = [];
-  var key;
+  const result = [];
+  let key;
 
   for (key in value) {
     result.push(key);
--- a/packages/remark-parse/lib/tokenize/text.js
+++ b/packages/remark-parse/lib/tokenize/text.js
@@ -6,23 +6,21 @@
  * @fileoverview Tokenise text.
  */
 
-'use strict';
-
-module.exports = text;
+export default text;
 
 /* Tokenise text. */
 function text(eat, value, silent) {
-  var self = this;
-  var methods;
-  var tokenizers;
-  var index;
-  var length;
-  var subvalue;
-  var position;
-  var tokenizer;
-  var name;
-  var min;
-  var now;
+  const self = this;
+  let methods;
+  let tokenizers;
+  let index;
+  let length;
+  let subvalue;
+  let position;
+  let tokenizer;
+  let name;
+  let min;
+  let now;
 
   /* istanbul ignore if - never used (yet) */
   if (silent) {
@@ -58,7 +56,7 @@ function text(eat, value, silent) {
   subvalue = value.slice(0, min);
   now = eat.now();
 
-  self.decode(subvalue, now, function (content, position, source) {
+  self.decode(subvalue, now, (content, position, source) => {
     eat(source || content)({
       type: 'text',
       value: content
--- a/packages/remark-parse/lib/tokenizer.js
+++ b/packages/remark-parse/lib/tokenizer.js
@@ -6,27 +6,23 @@
  * @fileoverview Markdown tokenizer.
  */
 
-'use strict';
+export default factory;
 
-module.exports = factory;
-
-var MERGEABLE_NODES = {
+const MERGEABLE_NODES = {
   text: mergeText,
   blockquote: mergeBlockquote
 };
 
 /* Check whether a node is mergeable with adjacent nodes. */
 function mergeable(node) {
-  var start;
-  var end;
+  let start;
+  let end;
 
   if (node.type !== 'text' || !node.position) {
     return true;
   }
-
   start = node.position.start;
   end = node.position.end;
-
   /* Only merge nodes which occupy the same size as their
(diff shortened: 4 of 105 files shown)


wooorm:
> Sorry for my late response.

Sorry for *my* late response! Thanks for working on this!

> Some of variables defined only once. So, xo is asking us to use const. If we do, the current convention, putting every variables on top, will be broken.

Do you have a code example that shows this? I don’t think I get what you’re saying.

> Introduce modern build(webpack/rollup?)/testing(Jest?)

Build:  The build isn’t very important. I use it to generate [releases](https://github.com/wooorm/remark/releases), but the npm package *is* important! Anyway, I’m fine with trying webpack/rollup/*, and would prefer the least amount of configuration.

Test: there’s a lot of tests. I’d say keep stuff at tape now and focus on updating the core packages!

> I need more feedback before we decide how should we refactor this.

Sure, like what?

> Also, I'm considering to rewrite whole code as a typescript or to prepare solid type definition.

Please don’t! Here’s some more [views of mine on preferring JS over ES2020/CoffeeScript/*](https://github.com/wooorm/remark/issues/252#issuecomment-297060057). I like JavaScript, and I like code people can read, I also like stable things. On the other hand, I’m pragmatic enough to try this out if it would mean a much smaller size, but only if this makes a difference!

Thanks again for working on this. Do let me know if you have questions.

:wave:

/cc @probablyup

Rokt33r:
> > I need more feedback before we decide how should we refactor this.
>
> Sure, like what?

The things what I've done and I'm going to do. Do they make sense?

### Use es6 module

```diff
-var start = require('unified-args');
-var extensions = require('markdown-extensions');
-var processor = require('remark');
-var proc = require('remark/package.json');
-var cli = require('./package.json');
+import start from 'unified-args';
+import extensions from 'markdown-extensions';
+import processor from 'remark';
+import proc from 'remark/package.json';
+import cli from './package.json';
```

### Use const and let rather than var

```diff
 function parse(options) {
-  var Local = unherit(Parser);
+  const Local = unherit(Parser);
    Local.prototype.options = xtend(Local.prototype.options, this.data('settings'), options);
    this.Parser = Local;
 }
```

### Enable jsnext(which is the option of xo)

https://github.com/wooorm/remark/pull/262/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L53

### Remove 'use strict'
```diff
-'use strict';
```

---

> > Some of variables defined only once. So, xo is asking us to use const. If we do, the current convention, putting every variables on top, will be broken.
>
> Do you have a code example that shows this? I don’t think I get what you’re saying.

I'll prepare the example soon.

wooorm:
@Rokt33r All makes sense to me! 👍

packages/remark-parse/lib/locate/escape.js
```diff
@@ -6,9 +6,7 @@
  * @fileoverview Locate an escape.
  */
 
-'use strict';
-
-module.exports = locate;
+export default locate;
 
 function locate(value, fromIndex) {
```
probablyup: for cases like this you can place the export keyword directly on locate, e.g.

`export default function locate()`

packages/remark-parse/lib/parser.js
```diff
@@ -5,17 +5,14 @@
  * @module remark:parse
  * @fileoverview Markdown parser.
  */
+import xtend from 'xtend';
```
probablyup: `lodash.merge` might make a good replacement. I wouldn't switch entirely to `Object.assign` since that requires polyfilling in many environments

packages/remark-parse/lib/tokenize/text.js
```diff
@@ -58,7 +56,7 @@ function text(eat, value, silent) {
   subvalue = value.slice(0, min);
   now = eat.now();
 
-  self.decode(subvalue, now, function (content, position, source) {
```
probablyup: I'd hold off on this sort of change unless it provides a benefit like avoiding `const that = this`. There is a perf cost to making a bunch of bound functions, though small.

packages/remark-parse/lib/tokenizer.js
```diff
@@ -6,27 +6,23 @@
  * @fileoverview Markdown tokenizer.
  */
 
-'use strict';
+export default factory;
```
probablyup: Remember that ESM exports are staticly defined. You shouldn't export before the object/function has been allocated or it will export `undefined` in strict loaders.

packages/remark-parse/lib/util/html.js
```diff
@@ -6,24 +6,22 @@
...
-var doubleQuoted = '"[^"]*"';
-var attributeValue = '(?:' + unquoted + '|' + singleQuoted + '|' + doubleQuoted + ')';
-var attribute = '(?:\\s+' + attributeName + '(?:\\s*=\\s*' + attributeValue + ')?)';
-var openTag = '<[A-Za-z][A-Za-z0-9\\-]*' + attribute + '*\\s*\\/?>';
-var closeTag = '<\\/[A-Za-z][A-Za-z0-9\\-]*\\s*>';
-var comment = '<!---->|<!--(?:-?[^>-])(?:-?[^-])*-->';
-var processing = '<[?].*?[?]>';
-var declaration = '<![A-Za-z]+\\s+[^>]*>';
-var cdata = '<!\\[CDATA\\[[\\s\\S]*?\\]\\]>';
+export default new RegExp('^(?:' + openTag + '|' + closeTag + ')');
```
probablyup: This is a different export behavior than the original, make sure you update all the callsites or perhaps just put it back to a named export?

packages/remark-parse/lib/parser.js
wooorm: `xtend` is very small and simple, and used a lot throughout all my plugins, so it probably doesn’t cost anything to keep it in, whereas I don’t use `lodash.merge` so it’ll probably add to the bundled size!

packages/remark-parse/lib/tokenizer.js
wooorm: @probablyup Wait, you can’t export a function that’s defined later?

packages/remark-parse/lib/parser.js
probablyup: Fair enough.

packages/remark-parse/lib/tokenizer.js
probablyup: To my knowledge, no. I tried googling again this morning but can't find any info either way. It was my understanding that the export keyword allocation is meant to be immutable... but I could be mistaken.

wooorm: Yeah I Googled too, but couldn’t find it out.

> It was my understanding that the export keyword allocation is meant to be immutable...

That’s right, but I’m unsure whether function hoisting is “mutable” though 😬

packages/remark-parse/lib/util/html.js
Rokt33r: I think I've made a mistake. I'll fix it and check any other places also.

packages/remark-parse/lib/parser.js
Rokt33r: Then, let's keep `xtend` and other helper functions.

Rokt33r:
I think I cannot afford to continue this for now. I need to figure out how should I deal with this.
So, I'm going to fork remark and try to rewrite in Typescript as an experiment. I'll be back here one or two months later.

Rokt33r:
Currently, I'm re-writing in Typescript. You can see the progress from the below link. It should be a complete overhaul against whole ecosystem of remark. Every package will be definitely typed and available for both es6 and commonjs module systems.

https://github.com/Rokt33r/typed-remark

Issue #259, opened 2017-06-04
Footnotes: two consecutive footnoteReference are parsed as linkReference


vhf:
Hey!

I think the behaviour I'm encountering is a bug. Here is some Markdown I `reparse`d with `footnotes: true` as only option:

```markdown
a[^foo]

a[^foo][^foo]

a[^foo][^foo][^foo][^foo]

a[^foo][^foo][^foo][^foo][^foo]

[^foo]: MyNote
```

I end up with this AST:

```
root[5] (1:1-11:1, 0-100)
├─ paragraph[2] (1:1-1:8, 0-7)
│  ├─ text: "a" (1:1-1:2, 0-1)
│  └─ footnoteReference (1:2-1:8, 1-7) [identifier="foo"]
├─ paragraph[2] (3:1-3:14, 9-22)
│  ├─ text: "a" (3:1-3:2, 9-10)
│  └─ linkReference[1] (3:2-3:14, 10-22) [identifier="^foo"][referenceType="full"]
│     └─ text: "^foo" (3:4-3:8, 12-16)
├─ paragraph[3] (5:1-5:26, 24-49)
│  ├─ text: "a" (5:1-5:2, 24-25)
│  ├─ linkReference[1] (5:2-5:14, 25-37) [identifier="^foo"][referenceType="full"]
│  │  └─ text: "^foo" (5:4-5:8, 27-31)
│  └─ linkReference[1] (5:14-5:26, 37-49) [identifier="^foo"][referenceType="full"]
│     └─ text: "^foo" (5:16-5:20, 39-43)
├─ paragraph[4] (7:1-7:32, 51-82)
│  ├─ text: "a" (7:1-7:2, 51-52)
│  ├─ linkReference[1] (7:2-7:14, 52-64) [identifier="^foo"][referenceType="full"]
│  │  └─ text: "^foo" (7:4-7:8, 54-58)
│  ├─ linkReference[1] (7:14-7:26, 64-76) [identifier="^foo"][referenceType="full"]
│  │  └─ text: "^foo" (7:16-7:20, 66-70)
│  └─ footnoteReference (7:26-7:32, 76-82) [identifier="foo"]
└─ footnoteDefinition[1] (9:1-9:15, 84-98) [identifier="foo"]
   └─ paragraph[1] (9:9-9:15, 92-98)
      └─ text: "MyNote" (9:9-9:15, 92-98)
```

[...]

PR #263, opened 2017-06-12
Allow consecutive footnotes on a node in `remark-parse`

vhf:
Here is an attempt to solve #259 

This is preliminary work, I don't expect tests to pass. It is unclear to me how this should behave for each combination of options, I don't know much about CommonMark for instance and GFM doesn't have footnotes support anyway.

Please tell me what you think is the best way to handle this, @wooorm :)

### DIFF
--- a/packages/remark-parse/lib/tokenize/reference.js
+++ b/packages/remark-parse/lib/tokenize/reference.js
@@ -122,7 +122,7 @@ function reference(eat, value, silent) {
 
   character = value.charAt(index);
 
-  if (character === C_BRACKET_OPEN) {
+  if ((self.options.footnotes && type !== T_FOOTNOTE) && character === C_BRACKET_OPEN) {
     identifier = '';
     queue += character;
     index++;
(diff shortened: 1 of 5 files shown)


wooorm:
Sorry for the delay, I’ve been busy IRL, and there’s more issues coming in than I can handle! 😬

Also: could you add the example [we talked about](https://github.com/wooorm/remark/issues/259#issuecomment-306217368) to the tests too? And make sure it generated what we came up with?

For the tests:

- you can create a test file in `input/`, as you’ve already done
- add a corresponding JSON file in `output/`.
- generate the fixture with `node script/regenerate-fixtures`
- run the tests with `npm run test-api`.

To make sure you have fixtures for other flavours, you can run `npm run test-api-extensive`, add the failing output test fixtures, regenerate, and run the tests again until everything’s okay.


Thanks so much for working on this, feel free to ping as much as needed to get me to help with this!

vhf:
Hey @wooorm , sorry I didn't put more time into this patch. :(

Thanks a lot for your fix!

wooorm:
Sure, no problem! I amended some of it but it didn’t take me long to fix, thanks for the initial work!

PR #264, opened 2017-06-14
Change require in defaults.js.

StevenXL:
Some setups don't work without the file extension. This will work with
all setups.

### DIFF
--- a/packages/remark-parse/lib/defaults.js
+++ b/packages/remark-parse/lib/defaults.js
@@ -16,6 +16,6 @@ module.exports = {
   commonmark: false,
   footnotes: false,
   pedantic: false,
-  blocks: require('./block-elements'),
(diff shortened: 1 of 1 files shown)


wooorm:
Hi Steven, thanks for reaching out, and thanks for PRing! 👋

Say I do that here, you’ll probably run into other modules doing similar things. I think this is a problem solved in webpack already, through the [`json-loader`](https://github.com/webpack/json-loader) (but I don‘t really know webpack). Then, you also need to add `.json` to [`resolve.extensions`](https://webpack.github.io/docs/configuration.html#resolve-extensions).

Hope this helps,
Titus

StevenXL:
@wooorm Hi. Thanks for the response. Yes, it is indeed solved through JSON loader. Thank you.

wooorm:
Great! :+1:

PR #277, opened 2017-07-10
Explicitly require `block-elements.json`

pkuczynski:
This fixes the error when Webpack3 is used and json is not part of the resolve config (see: https://github.com/grommet/grommet/issues/1018#issuecomment-256805982).

```
Module not found: Error: Can't resolve './block-elements' in '/Users/pkuczynski/Projects/shaker/node_modules/remark-parse/lib'
 @ ./node_modules/remark-parse/lib/defaults.js 19:10-37
 @ ./node_modules/remark-parse/lib/parser.js
 @ ./node_modules/remark-parse/index.js
```

### DIFF
--- a/packages/remark-parse/lib/defaults.js
+++ b/packages/remark-parse/lib/defaults.js
@@ -16,6 +16,6 @@ module.exports = {
   commonmark: false,
   footnotes: false,
   pedantic: false,
-  blocks: require('./block-elements'),
+  blocks: require('./block-elements.json'),
   breaks: false
 };


wooorm:
Cheers!

pkuczynski:
Would be great if you could release a new version, so I can request the same from `markdown-to-jsx` which I am actually using directly... Thx!

wooorm:
I get that, but sorry, no, I’m not going to release this just yet. In fact, I think this can be fixed by using the JSON loader, as that issue already mentioned. I also already declined some PRs to do this (e.g., https://github.com/wooorm/remark/issues/215), but now I caved because a lot of people run into this.

Anyway, I’ll probably cut a major release later this weak, due to some [other recent changes](https://github.com/wooorm/remark/commits/master)!

pkuczynski:
Yes, the `json-loader` works as a temporary solution. However, it should not be needed with latest webpack, as it has this loader built-in. The issue arises when one modifies `resolve` config.

Anyway, I can wait till next major upgrade I guess :)

PR #279, opened 2017-07-17
docs: add remark-embed-images plugin

dherges:
Hi @wooorm,

I created a new plugin that embeds images. It's still very very experimental and works for basic usages just


wooorm:
👍

PR #298, opened 2017-09-18
Add remark-shortcodes to plugins list

djm:
Hi,

Thanks for your work on Remark!

I've created a Markdown plugin which parses Wordpress/Hugo-like block-level shortcodes. It's a simple parser and just updates the AST with the appropriate details, the idea being that other plugins can then handle the transformation to something more meaningful like a template partial.

More here: https://www.npmjs.com/package/remark-shortcodes

Cheers!
Darian


wooorm:
Awesome! 👍

One suggestion for the readme: add a description of the new node type to [Not a transformer](https://github.com/djm/remark-shortcodes#not-a-transformer) in IDL format. Something like the [`footnoteReference` description](https://github.com/syntax-tree/mdast#footnotereference)?

djm:
@wooorm Aha, I was looking for the right type syntax. Will do. Cheers.

PR #300, opened 2017-09-22
Add support for turning off entity decoding

erquhart:
For Unified implementations that handle their own escaping, this is necessary for maintaining the decoded/encoded state of HTML entities in a markdown document.

### DIFF
--- a/packages/remark-parse/lib/defaults.js
+++ b/packages/remark-parse/lib/defaults.js
@@ -6,5 +6,6 @@ module.exports = {
   commonmark: false,
   footnotes: false,
   pedantic: false,
-  blocks: require('./block-elements.json')
+  blocks: require('./block-elements.json'),
+  decodeHtmlEntities: true
 };
--- a/packages/remark-parse/lib/tokenize/text.js
+++ b/packages/remark-parse/lib/tokenize/text.js
@@ -47,12 +47,20 @@ function text(eat, value, silent) {
   }
 
   subvalue = value.slice(0, min);
-  now = eat.now();
 
-  self.decode(subvalue, now, function (content, position, source) {
-    eat(source || content)({
-      type: 'text',
-      value: content
+  if (this.options.decodeHtmlEntities) {
+    now = eat.now();
+
+    self.decode(subvalue, now, function (content, position, source) {
+      eat(source || content)({
+        type: 'text',
+        value: content
+      });
     });
-  });
+  } else {
+    eat(subvalue)({
+      type: 'text',
+      value: subvalue
+    });
+  }
 }
--- a/packages/remark-parse/readme.md
+++ b/packages/remark-parse/readme.md
@@ -143,6 +143,17 @@ Pedantic mode (`boolean`, default: `false`) turns on:
 *   And pedantic mode removes less spaces in list-items (at most four,
     instead of the whole indent)
 
+##### `options.decodeHtmlEntities`
+
+```md
+&lt;div&gt;
+```
+
+Decode HTML entities (`boolean`, default: `true`)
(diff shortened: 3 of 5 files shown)


erquhart:
@wooorm let me know what you think, or if you'd like any changes made/tests added.

wooorm:
Hey @erquhart, sorry for the late reply! I’m super busy coordinating a class.

Anyway, about this feature, I don’t think it should be in core. In my opinion it should be a plugin instead. 

I’m thinking something like [`remark-breaks`](https://github.com/wooorm/remark-breaks), but then overwriting the `text` tokeniser.

Before digging into how that would work (I can advise a bit on how it could work), do you think that could work too?

erquhart:
@wooorm my concerns would be staying up to date with the core, and the effort it would take to understand how tokenizers are written. I've spent months on a comprehensive integration between Unified and Slate, and it's past time to release. I'm leaning toward forking if this can't be included, but really hoping to avoid that.

wooorm:
I’m not sure how forking would fix "staying up to date with core". You underestimate the time involve to maintain a fork, and overestimate the time it takes to create a plugin.

You‘re also underestimating the time involved in maintaining code on one hand, if it were included in core, and overestimate it in a plugin, if the community were to maintain it. It’s the same code, and takes the same amount of time to maintain. Question is who spends that time.

Aside: you are of course free to fork. [I grant you that right](https://github.com/wooorm/remark/blob/master/LICENSE). But, I’ve seen a lot of forks of remark, and they all grow stale in the end. Maybe your fork is special though?

> Perhaps writing the plugin would be simpler than I expect.

Yes, it would be. As I previously asked (“I can advise a bit on how it could work”), can I help you with that?

erquhart:
I definitely didn't mean to be contentious - apologies if I came off that way. I have no business communicating with the outside world before coffee. I love Unified! It's a fantastic ecosystem and you've done amazing work; the idea of forking isn't meant to detract, only for expedience. The advantage of a fork is that I can get this released now and, while it would grow stale, it would do so uniformly. As I stated, it's not the ideal approach, for sure. I've written ten Unified plugins within the Netlify CMS project so far, but have avoided digging into internals 😅 

I'll definitely take any assistance you can offer!

wooorm:
Alright, thanks for that. I get it though, we both have important goals here; but still I just wanted to voice my opinion too. I appreciate the kind words, and I appreciate you choosing unified and friends!

Your plugin (not sure of a name), would essentially be similar to [`remark-breaks`](https://github.com/wooorm/remark-breaks/blob/master/index.js) (so ± 50 lines). It would **overwrite** the `text` [`inlineTokenizer`](https://github.com/wooorm/remark/tree/master/packages/remark-parse#parserinlinetokenizers).

So you’d set essentially the new [`text`](https://github.com/erquhart/remark/blob/09e9abcefa556f8f1c0c6f4f2589c0c223f4a6ab/packages/remark-parse/lib/tokenize/text.js#L5-L66) function, but without the standard stuff that’s currently happening (as in, `decodeHtmlEntities: true`).

The [big part of `text`](https://github.com/erquhart/remark/blob/09e9abcefa556f8f1c0c6f4f2589c0c223f4a6ab/packages/remark-parse/lib/tokenize/text.js#L23-L47) is actually something we could externalise in core to another function, so it’s easier to use in your plugin. So if you’d like to use less code in your plugin, making it easier to maintain, we could do that too!

erquhart:
Externalizing the core text functionality would be a huge help! I'll just port what I need for now and update when that's ready.

wooorm:
Hey it’s been a while, sorry about that! 😬
I’m going to close this PR for now, could you raise a new issue about externalising code, if you still need it?

PR #318, opened 2018-01-14
fix(url): disallow `nonTerminated` entities

### LINKED ISSUE #307: Remark change &reg into ® in URL
At each instance of the `& reg` element, remark transforms it to ®, even in URLs.

Input is :
```markdown
[example 1](http://www.example.com?param=1&region=here)
http://www.example.com?param=1&region=here
`http://www.example.com?param=1&region=here`
```

Output is :
```markdown
[example 1](http://www.example.com?param=1®ion=here)
<http://www.example.com?param=1®ion=here>
`http://www.example.com?param=1&region=here`
```

I think it should have an exception on URL formatting.

ikatyang:
Fixes #307 

Include `url`s in:
- link
- image
- autolink
- definition


wooorm:
Merged! I’ll cut a release soon, thanks for waiting!

PR #319, opened 2018-01-14
fix(list): interrupted by `atxHeading`

### LINKED ISSUE #313: A heading immediately after a list element is parsed as an item of that list
Originally reported to prettier: https://github.com/prettier/prettier/issues/3247

See http://astexplorer.net/#/gist/e3eaf84774057db35a17b3065aa3f8a3/69aa50de7bc313fad863946ebe318aa88d288afa

ikatyang:
Fixes #313 

I'm not sure if it should be applied to the `default` mode, so I did not enable it by default.


wooorm:
Sweet! I instead went with `pedantic: false`, because it does not work in [pedantic mode](https://daringfireball.net/projects/markdown/dingus) specifically, but does in the rest!

PR #328, opened 2018-03-09
fixes #327 - trailing spaces are not required in empty ordered list

### LINKED ISSUE #327: [CommonMark] trailing spaces are not required in empty ordered list
### Subject of the issue

`commonmark: true`

```md
1.
2.
3.
```

### Steps to reproduce

[astexplorer](http://astexplorer.net/#/gist/548c2642132c8284e88b3291726ee408/80875274e506492bae6f81949c25762ba5f80855)

### Expected behaviour

Same as [CommonMark playground](http://spec.commonmark.org/dingus/?text=1.%0A2.%0A3.%0A):

```json
{
  "type": "list",
  "ordered": true,
  "start": 1,
  "loose": false,
  "children": [
    {
      "type": "listItem",
      "loose": false,
      "checked": null,
      "children": []
    },
    {
      "type": "listItem",
      "loose": false,
      "checked": null,
      "children": []
    },
    {
      "type": "listItem",
      "loose": false,
      "checked": null,
      "children": []
    }
  ]
}
```

### Actual behaviour

```json
{
  "type": "paragraph",
  "children": [
    {
      "type": "text",
      "value": "1.\n2.\n3."
    }
  ]
}
```

mlrawlings:
Fixes #327 (and https://github.com/prettier/prettier/issues/4122): 
With `commonmark:true`, an ordered list should not require spaces after the list marker.
[commonmark.js dingus](http://spec.commonmark.org/dingus/?text=1.%0A2.%0A3.%0A)

**In**:
```
1.
2.
3.
```
**Out**:
1.
2.
3.


packages/remark-parse/lib/tokenize/list.js
```diff
@@ -139,7 +141,13 @@ function list(eat, value, silent) {
 
   character = value.charAt(++index);
 
-  if (character !== C_SPACE && character !== C_TAB) {
+  delimiters = {};
+  delimiters[C_SPACE] = true;
+  delimiters[C_TAB] = true;
+  delimiters[C_NEWLINE] = commonmark;
+  delimiters[C_EMPTY] = commonmark;
```
ChristianMurphy: With delimiters empty being checked, how will

```markdown
1.2.3.
```

be parsed?

mlrawlings: It will just be text.  

If you look on line [142](https://github.com/remarkjs/remark/pull/328/files/f3c06ab664be596d89729c7cd50fbe12626d29af#diff-0b27a32611df78dd18ed7a0b170ec571R142), you can see we grab the next `character` to see if it is a valid delimiter.  This is the only character compared against the delimiters and the only time it would be empty is when there is no next character (the end of the file has been reached).

So in the case of `1.2.3.`, `2` would be the next character after matching the first list-item-looking-thing and that's not a valid delimiter, so we'd early return and it would not be parsed as a list.

PR #331, opened 2018-04-24
remark-parse: parse link/img's attributes

arobase-che:
Hi,

This pull request add a parser to link and image elements.

Images : 
```markdown
![alt](img){attrs} / ![alt](img){ height=50 }
```

Link   :
```markdown
[Hot babe with computer](https://rms.sexy){rel="external"}
```
At the moment it aims is to be used with remark-rehype to add attributes :
```html
<img alt="alt" src="img" height="50">

<a href="https://rms.sexy" rel="external">Hot babe with computer</a>
```

Not enabled when gfm is specified.

I'm open to any review. I think that this PL should be added as an option. ;)

Have a good day

### DIFF
--- a/packages/remark-parse/lib/tokenize/link.js
+++ b/packages/remark-parse/lib/tokenize/link.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var whitespace = require('is-whitespace-character');
+var parseAttr = require('md-attr-parser');
 var locate = require('../locate/link');
 
 module.exports = link;
@@ -42,6 +43,7 @@ function link(eat, value, silent) {
   var pedantic = self.options.pedantic;
   var commonmark = self.options.commonmark;
   var gfm = self.options.gfm;
+  var parsedAttr;
   var closed;
   var count;
   var opening;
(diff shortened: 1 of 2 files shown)


wooorm:
This looks cool!

This is not a standard function of markdown though, so it shouldn’t be in core. Rather, it should be an external plugin. Most plugins work on the tree, but this should be something that integrates with the parser. There are some other plugins like that already, such as [remark-breaks](https://github.com/remarkjs/remark-breaks), [remark-frontmatter](https://github.com/remarkjs/remark-frontmatter), or [remark-gemoji](https://github.com/remarkjs/remark-gemoji).

So, I think it’s great, but shouldn’t be in core!

arobase-che:
I totaly agree with you but it was the easier way to do it  :grin: 
Commonmark should ( ... we know that's not for tomorrow) spec it.

I will see how other plugins do it. It's look complicated :( 
See you soon ! ;)

sasivarnan:
@arobase-che I will also be interested to have this feature as it comes handy in many situations.
Working on a plugin for this?

sasivarnan:
@arobase-che Great to hear. 👍 
Looking forward to exploring it. 🥇

wooorm:
Awesome!

arobase-che:
Ready  💕

[Github :octocat:](https://github.com/arobase-che/remark-attr)

[npm](https://www.npmjs.com/remark-attr)

PR #344, opened 2018-07-24
Fix: pedantic remark-stringify can produce markdown that cannot be pedantically parsed

Hamms:
The specific problem this PR addresses is with underscores in emphasis. In pedantic mode, underscores in emphasis are only allowed if the emphasis is created with asterisks, not underscores:

-  `*alpha_bravo*` compiles to `<em>alpha_bravo</em>`
-  `_alpha_bravo_` compiles to `<em>alpha</em>bravo_`

Unfortunately, `remark-stringify` as it currently stands will stringify both inputs to `_alpha_bravo_`, meaning that the result of the chain:

    remark-parser -> remark-stringify -> remark-parser -> remark-to-rehype -> rehype-stringify

will be different from the result of the chain:

    remark-parser -> remark-to-rehype -> rehype-stringify

in pedantic mode.

The goal of this PR is to ensure that those two chains have identical outputs; or, put another way, to ensure that `remark-stringify` is not a destructive process, even in pedantic mode.

### DIFF
--- a/packages/remark-stringify/lib/visitors/emphasis.js
+++ b/packages/remark-stringify/lib/visitors/emphasis.js
@@ -19,10 +19,15 @@ function emphasis(node) {
   var marker = this.options.emphasis;
   var content = this.all(node).join('');
 
-  if (this.options.pedantic) {
-    if (marker === '_' && content.indexOf(marker) !== -1) {
-      marker = '*';
-    }
+  /* When in pedantic mode, prevent using underscore as the marker when
+   * there are underscores in the content.
+   */
+  if (
+    this.options.pedantic &&
+    marker === '_' &&
+    content.indexOf(marker) !== -1
+  ) {
+    marker = '*';
   }
 
   return marker + content + marker;
--- a/packages/remark-stringify/lib/visitors/emphasis.js
+++ b/packages/remark-stringify/lib/visitors/emphasis.js
@@ -9,8 +9,21 @@ module.exports = emphasis;
  * asterisk (`'*'`):
  *
  *     *foo*
+ *
+ * In `pedantic` mode, text which itself contains an underscore
+ * will cause the marker to default to an asterisk instead:
+ *
+ *     *foo_bar*
  */
 function emphasis(node) {
   var marker = this.options.emphasis;
-  return marker + this.all(node).join('') + marker;
+  var content = this.all(node).join('');
+
+  if (this.options.pedantic) {
+    if (marker === '_' && content.indexOf(marker) !== -1) {
+      marker = '*';
+    }
+  }
+
+  return marker + content + marker;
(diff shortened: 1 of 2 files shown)


packages/remark-stringify/lib/visitors/emphasis.js
```diff
@@ -9,8 +9,21 @@ module.exports = emphasis;
...
+ *
+ *     *foo_bar*
  */
 function emphasis(node) {
   var marker = this.options.emphasis;
-  return marker + this.all(node).join('') + marker;
+  var content = this.all(node).join('');
+
+  if (this.options.pedantic) {
+    if (marker === '_' && content.indexOf(marker) !== -1) {
```
ChristianMurphy: this conditional can be combined with the one on line 22.

packages/remark-stringify/lib/visitors/emphasis.js
```diff
@@ -9,8 +9,21 @@ module.exports = emphasis;
...
+      marker = '*';
```
ChristianMurphy: I'm not sure about overriding the `marker` setting, @wooorm what do you think?

Hamms: Yeah, I wasn't sure how to feel about that, either.

I considered trying to do this fallback if and only if the `emphasis` option had been left set to the default, and skipping it if it had been manually set to `'_'`, but I wasn't sure how to feel about that, either.

wooorm:
Well, pedantic mode is pretty broken. I think the simplest course for you would be to drop it instead of trying to fix it! We could also add docs for pedantic mode to say that `*` should be used as a marker, or even enforce it?

With regards to fixing this, I’d like to see some tests of other stuff embedded in emphasis as well (think URLs, images, HTML, etc etc.).

But other than that, I think ignoring the marker option and going with `_` could work.

Hamms:
> Well, pedantic mode is pretty broken. 

Would it be possible for the documentation to be updated to warn about that? I definitely did not know that was the consensus when I picked up this project.

I do intend to eventually move away from pedantic mode, but I'm in a transitional period between markdown standards now so I gotta do what I gotta do.

> We could also add docs for pedantic mode to say that * should be used as a marker, or even enforce it?

That would definitely be a lot less weird than pedantic mode _sometimes_ switching up the default marker on you; I'd be happy to update this PR with that functionality if y'all are on board.

Hamms:
@ChristianMurphy @wooorm do you have an estimate on when this can get merged? I'd love to get this functionality into my project.

ChristianMurphy:
Barring objections I think it could be merged today.

@wooorm question, squash and merge is the only option, is that intentional?

Hamms:
Thanks!

Issue #342, opened 2018-07-21
Support for fenced code block info string


arobase-che:
Hi,

I'm looking for a support for fenced code block info string. It's not a regular markdown but it's pretty commun and both [GFM](https://github.github.com/gfm/#info-string) and [CommonMark](https://spec.commonmark.org/0.28/#info-string) spec it.

If i implement it and make a pull request, will it be merged or should i make a plugin ?

Thanks ! :wink:

ChristianMurphy:
Since it is part of the common mark spec, I'd personally lean toward including this in remark directly.

---

It'd be good to get the specification example as a test

```markdown
~~~~    ruby startline=3 $%@#$
def foo(x)
  return 3
end
~~~~~~~
```

https://spec.commonmark.org/0.28/#example-112

---

and also would be good to get the MDAST spec updated with the new attribute
https://github.com/syntax-tree/mdast#code

arobase-che:
Arf, in fact ... While discussing with @hypercube he remarked to me that the info string is the whole string. In you example "ruby starline=3 $%@#$". Which is exactly the value of the "lang" attribute.

So "lang" **is** actually the info string. Now "lang" is the info string, but it's used by other plugins as the first word of the info string as it's name suggests it is.

I will make a PR soon fix it.

Hypercubed:
The name is @hypercubed...  :)

Anyway, yeah, from the spec:

> The content of a code fence is treated as literal text, not parsed as inlines. The first word of the info string is typically used to specify the language of the code sample, and rendered in the class attribute of the code tag. However, this spec does not mandate any particular treatment of the info string.

So, if you want to keep backword compat, the "info string" should be the whole string and `lang` should be the first word of teh info string.  This might be breating for existing code that look for the full info string in `lang`...  (it will break me, but for the greater good?).

wooorm:
@Hypercubed Yeah I was thinking something along those lines, a `code.infoString` being the whole thing, `lang` being the first “word”? It’s breaking, but more fixing.

@arobase-che Thanks for the issue! Sorry for the slow response. I’m also happy to see how you worked “remark” in your comment (“...he remarked to me...”) 😛

arobase-che:
needs pr you said ?

Ok let's go. Where are the tests about remark-parse ? I only found [that file](https://github.com/remarkjs/remark/blob/02297a5dea290751a0bfd766b4154ad403307b7e/test/remark-parse.js).

ChristianMurphy:
[...]

PR #345, opened 2018-07-28
Support for fenced code block info string

arobase-che:
Hi \o

The discussion is on the corresponding issue remarkjs/remark#342

Spoiler: It's about info string ;)

💘

### DIFF
--- a/packages/remark-parse/lib/tokenize/code-fenced.js
+++ b/packages/remark-parse/lib/tokenize/code-fenced.js
@@ -230,7 +230,8 @@ function fencedCode(eat, value, silent) {
 
   return eat(subvalue)({
     type: 'code',
-    lang: flag || null,
+    lang: flag.replace(/[\s].*/, '') || null,
+    info: flag.replace(/^[^\s]+/, '').trim() || null,
     value: trim(exdentedContent)
   });
 }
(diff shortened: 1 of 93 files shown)


wooorm:
Build failed for some reason, I restarted it and now it’s green again!

arobase-che:
¯\_(ツ)_/¯

wooorm:
@arobase-che Could you update the code here to match syntax-tree/mdast#22?

arobase-che:
Humm, it's already done. See [that comment](https://github.com/syntax-tree/mdast/pull/22#issuecomment-409415274). I updated the two the same day (night ...).

wooorm:
@arobase-che Sorry, missed it!  Soooo, I was in the process of merging this, but hmm, from what I read in the commonmark spec they call the whole thing the info string (I still like info instead of info string). However, as we’re removing the lang from it, I’m not sure `info` is still fitting. Would it make sense to use `meta` instead? I’m not sure where I saw it today, but I like it:

```
~~~js foo bar baz
   ^^^^^^^^^^^^^^ info
      ^^^^^^^^^^^ meta
   ^^ lang
console.log('!')
~~~
```

/CC @Hypercubed

ChristopherBiscardi:
Would love to see this get merged as we're getting requests for support in downstream projects ([gatsby-mdx](https://github.com/ChristopherBiscardi/gatsby-mdx/issues/76) and [mdx](https://github.com/mdx-js/mdx/issues/141)).

Is there anything I can do to help move this along? It looks like it might just need final review.

wooorm:
Landed!

PR #346, opened 2018-08-08
Add reference property on full references when different from identif…

rubys:
…ier.

* Change parse to add this property when needed to reconstruct the original
  input should the reference turn out to be undefined.
* Change parse test to handle this case
* Change stringify to prefer the reference property, if present, over
  the identifier property to produce markdown source.
* Add round-tripping tests for all undefined references.

### DIFF
--- a/test/remark-stringify.js
+++ b/test/remark-stringify.js
@@ -264,7 +264,7 @@ test('remark().stringify(ast, file)', function (t) {
       ['capitalized image references - full', '![alpha][Bravo]'],
       ['capitalized image references - collapsed', '![Bravo][]'],
       ['capitalized image references - shortcut', '![Bravo]'],
-      ['capitalized footnote references', '[^Alpha]'],
+      ['capitalized footnote references', '[^Alpha]']
     ];
 
     st.plan(tests.length);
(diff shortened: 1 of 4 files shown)


test/remark-stringify.js
```diff
@@ -255,6 +255,30 @@ test('remark().stringify(ast, file)', function (t) {
...
+  t.test('undefined references should fallback to original', function (st) {
+    /* Data-driven tests in the format: [name, input, expected] */
+    var tests = [
+      ['capitalized link references - full', '[alpha][Bravo]'],
+      ['capitalized link references - collapsed', '[Bravo][]'],
+      ['capitalized link references - shortcut', '[Bravo]'],
+      ['capitalized image references - full', '![alpha][Bravo]'],
+      ['capitalized image references - collapsed', '![Bravo][]'],
+      ['capitalized image references - shortcut', '![Bravo]'],
+      ['capitalized footnote references', '[^Alpha]'],
```
BridgeAR: Seems like the linter complains about the trailing comma.

BridgeAR:
Ping @wooorm

wooorm:
Superseded by GH-363, which takes this PR but uses a different name in accordance with mdast and adds labels to all associations.

PR #348, opened 2018-08-14
[CommonMark] Fix: paragraph split detection in commonmark mode

Hamms:
In commonmark, paragraphs can be split by blank lines _even if those lines contain whitespace_. ([example](https://spec.commonmark.org/dingus/?text=This%20is%20a%20paragraph%0A%20%20%20%20%0Aand%20this%20is%20a%20new%20paragraph))

Currently, in remark, paragraphs will be split by blank lines _only if those lines contain fewer than four characters worth of whitespace_, because the code to detect following indented content checks for an indentation but doesn't actually check for _content_.

The simple fix is to make sure that there is content following the indentation before absorbing the next paragraph.

I've added a new test fixture to cover this case; please let me know if that's not the way you'd prefer to test this change!

### DIFF
--- a/test/fixtures/input/paragraphs-and-indentation.text
+++ b/test/fixtures/input/paragraphs-and-indentation.text
@@ -6,6 +6,10 @@ This is a paragraph
 This is a paragraph
    and this is further text
 
+This is a paragraph
+    
+and this is a new paragraph
+
 This is a paragraph with some asterisks
     ***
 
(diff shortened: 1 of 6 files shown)


test/fixtures/input/paragraphs-and-indentation.text
```diff
@@ -6,6 +6,10 @@ This is a paragraph
 This is a paragraph
    and this is further text
 
+This is a paragraph
+
```
Hamms: note the whitespace here

Hamms:
@ChristianMurphy and/or @wooorm could I get a review when y'all get a chance?

PR #355, opened 2018-09-20
[Commonmark] Disallow whitespace between link text and link label in reference links

Hamms:
See [this section in the commonmark spec](https://spec.commonmark.org/0.28/#example-513) for context; in short, commonmark deviates from the original markdown spec in regards to to whitespace between the link text and the link label in an effort to improve support for shortcut reference links. Remark sticks to the original definition, even in commonmark mode.

For example:

```md
[foo] [bar]

[bar]: /url "title"
```

In Commonmark:

```html
<p>[foo] <a href="/url" `title="title">bar</a></p>
```

In Remark:

```html
<p><a href="/url" title="title">foo</a></p>
```

This PR changes the reflink parser only in commonmark mode, to match the commonmark spec.

Note that the nature of the change is most clear when [reviewing with whitespace changed ignored](https://github.com/remarkjs/remark/pull/355/files?w=1)

### DIFF
--- a/packages/remark-parse/lib/tokenize/reference.js
+++ b/packages/remark-parse/lib/tokenize/reference.js
@@ -20,6 +20,7 @@ var C_BRACKET_CLOSE = ']';
 
 function reference(eat, value, silent) {
   var self = this;
+  var commonmark = self.options.commonmark;
   var character = value.charAt(0);
   var index = 0;
   var length = value.length;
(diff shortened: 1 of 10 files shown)


Hamms:
@wooorm could you take a look at this when you get a chance?

Hamms:
Thanks!

PR #360, opened 2018-10-03
Improve "Turning off a tokenizer" in readme

bjudson:
* Recommend replacing tokenizer with noop, rather than deleting
* Link to plugin for disabling tokenizers

### DIFF
--- a/packages/remark-parse/readme.md
+++ b/packages/remark-parse/readme.md
@@ -390,15 +390,17 @@ The given `node`.
 ### Turning off a tokenizer
 
 In rare situations, you may want to turn off a tokenizer to avoid parsing
-that syntactic feature.  This can be done by deleting the tokenizer from
+that syntactic feature.  This can be done by replacing the tokenizer from
 your Parser’s `blockTokenizers` (or `blockMethods`) or `inlineTokenizers`
 (or `inlineMethods`).
 
 The following example turns off indented code blocks:
 
 ```js
-delete remarkParse.Parser.prototype.blockTokenizers.indentedCode;
+remarkParse.Parser.prototype.blockTokenizers.indentedCode = function () { return true };
 ```
+
+Preferably, just use [this plugin](https://github.com/zestedesavoir/zmarkdown/tree/master/packages/remark-disable-tokenizers).
 
 ## License
 


packages/remark-parse/readme.md
```diff
@@ -390,16 +390,18 @@ The given `node`.
...
-that syntactic feature.  This can be done by deleting the tokenizer from
+that syntactic feature.  This can be done by replacing the tokenizer from
 your Parser’s `blockTokenizers` (or `blockMethods`) or `inlineTokenizers`
 (or `inlineMethods`).
 
 The following example turns off indented code blocks:
 
 ```js
-delete remarkParse.Parser.prototype.blockTokenizers.indentedCode;
+remarkParse.Parser.prototype.blockTokenizers.indentedCode = () => true;
```
wooorm: Nice! Could you make use `function () {}` instead of an arrow function? I believe more people will then get it then.
P.S. Is the `true` needed? Doesn’t that throw errors, and conflict with silent mode? 🤔

bjudson: It doesn't throw errors for me (this noop function was lifted from the `remark-disable-tokenizers` plugin). The results seem preferable to me when it returns `true`, though I haven't done extensive testing:
```
parse.Parser.prototype.blockTokenizers.list = () => null;

unified()
  .use(parse)
  .use(html)
  .process("> * a\n* b\n* _c_\n")
  .then(value => console.log(value));

// output
<p>> <em> a</em> b* <em>c</em></p>
```
vs
```
parse.Parser.prototype.blockTokenizers.list = () => true;

unified()
  .use(parse)
  .use(html)
  .process("> * a\n* b\n* _c_\n")
  .then(value => console.log(value));

// output
<p>> * a</p><p>* b</p><p>* <em>c</em></p>
```

wooorm: Could you just check how it works with `false`? If that’s the same, or doesn’t work, I’m up for this 👍

bjudson: In my tests `false` and `null` produce the same result. (Also force-pushed change from arrow function to standard function.)

wooorm: OK, so only try works. Well let’s go with that, then!

Issue #350, opened 2018-08-29
[CommonMark] false positive for list.loose caused by trailing newline with whitespaces


ikatyang:
### Subject of the issue

`commonmark: true`

```md
- 123
  
```

(Note that there're two whitespaces in the last line.)

### Steps to reproduce

[astexplorer](http://astexplorer.net/#/gist/14e34ff614ea7abcbfc5d917e3351881/c1d57d382b5a13ac4c470d480272648f58e93490)

### Expected behaviour

Same as [CommonMark playground](https://spec.commonmark.org/dingus/?text=-%20123%0A%20%20):

```json
{
  "type": "root",
  "children": [
    {
      "type": "list",
      "ordered": false,
      "start": null,
      "loose": false,
      "children": [
        {
          "type": "listItem",
          "loose": false,
          "checked": null,
          "children": [
            {
              "type": "paragraph",
              "children": [{ "type": "text", "value": "123" }]
            }
          ]
        }
      ]
    }
  ]
}
```

### Actual behaviour

```diff
 {
   "type": "root",
   "children": [
     {
       "type": "list",
       "ordered": false,
       "start": null,
-      "loose": false,
+      "loose": true,
       "children": [
         {
           "type": "listItem",
-          "loose": false,
+          "loose": true,
           "checked": null,
           "children": [
             {
               "type": "paragraph",
               "children": [{ "type": "text", "value": "123" }]
             }
           ]
         }
       ]
     }
   ]
 }
```

Original post: prettier/prettier#5021

wooorm:
Probably related to GH-349!

Issue #358, opened 2018-09-29
Broken Stringify a list item


artalar:
### Subject of the issue

Broken Stringify a list item

### Your environment

*   windows 7
*   Names and version of required packages
*   remark@9.0.0, remark-stringify@5.0.0

### Steps to reproduce

```js
if (node.type === 'list') remark().stringify(node)
```

https://codesandbox.io/s/81kw44yowj

### Expected behaviour

return the string

### Actual behaviour

`TypeError: Cannot read property 'length' of undefined`

[this ](https://github.com/remarkjs/remark/blob/master/packages/remark-stringify/lib/macro/one.js#L20) `listItem` call paste not enough arguments

wooorm:
@artalar What output would you expect? How does one list-item look?

Also, why not:

```js
if (node.type === 'list') remark().stringify(node)
```

artalar:
@wooorm i was attaching the link to reproduce (https://codesandbox.io/s/81kw44yowj)

```js
const remark = require("remark");

const nodes = remark().parse(
  `
- 1
- 2
`
);

// expected `result` to be  `['1', '2']`
// currently `TypeError: Cannot read property 'length' of undefined`
const result = nodes.children[0].children.map(remark().stringify);
```

> `Also, why not:`

Becouse i need to separete every list item text

wooorm:
Well, the list holds information on how to compile the list-items, e.g., which bullet to use.
How could we [create bullets](https://github.com/remarkjs/remark/blob/ddc948a779d3078b5119d333a261745c8874f1ff/packages/remark-stringify/lib/visitors/list-item.js#L28) if we don’t have this information 🤔

artalar:
@wooorm [how I supose] list-item **value** doesn't know about _list_ context and doesn't need a bullet, because, a bullet - attribute of a list-item, not of these value.

artalar:
@wooorm I don't get it. So why I have the error?

How I can get string (raw) value of list-item?

artalar:
@wooorm how can I passing the contents of a list-item to `stringify`? Can you add the code snippet?

wooorm:
say `node` was a listItem, you can pass the contents with a `node.children.map(stringify)`.
Also note that https://github.com/remarkjs/remark/pull/364 is fixing this.

artalar:
@wooorm get it, thanks, it works!

But with _blockquote_ i have got new issue =\

(https://codesandbox.io/s/81kw44yowj)
[...]

Issue #332, opened 2018-04-25
[CommonMark] multiple empty lines between `listItem`s cause separate `list`s


ikatyang:
### Subject of the issue

`commonmark: true`

```md
* hello1

* hello2


* hello3

* hello4


* hello5

* hello6
```

### Steps to reproduce

[astexplorer](http://astexplorer.net/#/gist/1997547f6499eddace4fc67567755ca3/f3419b39b69169388ca3269b57808e5f72be80b8)

### Expected behaviour

Same as [CommonMark playground](http://spec.commonmark.org/dingus/?text=*%20hello1%0A%0A*%20hello2%0A%0A%0A*%20hello3%0A%0A*%20hello4%0A%0A%0A*%20hello5%0A%0A*%20hello6):

```json
[
    {
        "type": "list",
        "ordered": false,
        "start": null,
        "loose": true,
        "children": [
            {
                "type": "listItem",
                "loose": true,
                "checked": null,
                "children": [
                    {
                        "type": "paragraph",
                        "children": [
                            {
                                "type": "text",
                                "value": "hello1"
                            }
                        ]
                    }
                ]
            },
            {
                "type": "listItem",
                "loose": true,
                "checked": null,
                "children": [
                    {
                        "type": "paragraph",
                        "children": [
                            {
                                "type": "text",
                                "value": "hello2"
                            }
                        ]
                    }
                ]
            },
            {
                "type": "listItem",
                "loose": true,
                "checked": null,
                "children": [
                    {
                        "type": "paragraph",
                        "children": [
                            {
                                "type": "text",
                                "value": "hello3"
                            }
                        ]
                    }
                ]
            },
            {
                "type": "listItem",
                "loose": true,
                "checked": null,
                "children": [
                    {
                        "type": "paragraph",
                        "children": [
                            {
                                "type": "text",
                                "value": "hello4"
                            }
                        ]
                    }
                ]
            },
            {
                "type": "listItem",
                "loose": true,
                "checked": null,
                "children": [
                    {
                        "type": "paragraph",
                        "children": [
                            {
                                "type": "text",
                                "value": "hello5"
                            }
                        ]
                    }
                ]
            },
            {
                "type": "listItem",
                "loose": false,
                "checked": null,
                "children": [
                    {
                        "type": "paragraph",
                        "children": [
                            {
                                "type": "text",
                                "value": "hello6"
                            }
                        ]
                    }
                ]
            }
        ]
    }
]
```

### Actual behaviour

```json
[
[...]

PR #364, opened 2018-10-06
Rework lists (request for comments)

wooorm:
Hi folks! 👋

After some thinking about lists, such as syntax-tree/mdast-util-to-hast#23 and #349, and the changes in syntax-tree/mdast#24, I reworked the lists as they are currently handled by remark-parse and remark-stringify. Below you can see the changes, but the most importent one is that I propose to remove `loose` on `list` and on `listItem`, instead replacing them by `spread`.

This PR additionally closes GH-350, GH-332, and GH-358.

/cc Could you review this @Hamms, as you spend some time thinking about this as well, and @ikatyang, as this will affect prettier?

#### Spread

So, previously we had loose. Any list item that was followed by a blank line between in and its next sibling, or if it had a blank line between any of its children, would result in that list item to receive `loose: true`, and otherwise `loose: false`.
A list would have `loose: true` if any of its children had `loose: true`, and `loose: false` otherwise.

This was based on the `loose` definition of CommonMark, which means it is meant for HTML (namely, in a loose list, all `<p>` tags are output, whereas in a `tight` list they are not).

This was okay, but wasn’t really useful for either markdown (`remark-stringify`) *or* html (`mdast-util-to-hast`), as both need different mechanisms (remark needs to know where to place blank lines, in the list, and in the items; `mdast-util-to-hast` needs to know whether to output `<p>` tags).

In this new solution, we break from the commonmark definition, by introducing a new term:
* Lists have `spread: true` if there is a blank line *between* any of its children (not in them), in which case `remark-stringify` will join all children with blank lines between them
* Items have `spread: true` if there is a blank line *between* any of its children (not in them), in which case `remark-stringify` will join all children with blank lines between them

This decouples the two, which makes for a simpler AST, but it comes with the downside that `mdast-util-to-hast`, when stringifying an item, needs to calculate whether its parent list or any of its siblings has `spread: true`, to see whether it needs to generate `<p>` tags.

This is good though, because the tree stable. Say we injected a new item into a list, that would previously create an unstable tree (stringifying it and parsing it again, would result in a different tree).

#### Changes

remark-parse:

* Remove `loose` fields from `list`, `listItem`
* Add support for empty list-items in any non-pedantic mode
* Add `spread` to `list`, to signal that its items should have blank lines between them, set to true if any blank line exists between items
* Add `spread` to `listItem`, to signal that its children should have blank lines between them
* Fix bug where two lists, if they are both ordered or both not ordered, was seen in commonmark as one list

remark-stringify:

* Add support for missing `ordered` field
* Add support for missing `start` field
* Add support for `list.spread` (default: false)
* Add support for `listItem.spread` (default: true)
* Fix bug where two lists, if they are both ordered or both not ordered, and in commonmark, were not properly stringified, resulting in a new parse seeing them as one list
* Fix bug where a list followed by an indented code block, in commonmark, was not correctly stringified, resulting in a new parse seeing the code as content of the list
* Remove trailing white-space on empty list-items
* Add support for stringifying list-items without parent


ikatyang:
This looks great! 👍

But I think we still need the old `loose` field (maybe added by another plugin) to identify if two ASTs are the same from the perspective of CommonMark.

wooorm:
@ikatyang Of course! I think we can do it as some tiny utility!

wooorm:
@Hamms Thanks for reviewing!

> I love the idea of decoupling the two, and it makes sense that that would be necessary to best serve both sets of interests.

Perfect!

> If so, it's not clear to me that it's best to rename the field from loose to spread; can you explain a bit more what your thinking is there?

I meant a break away from *commonmark terminology*, and I think if these properties do not really do what CommonMark calls “loose”, then it makes more sense to have a different word for it!

wooorm:
> Agreed! I'm just having a hard time convincing myself one way or the other whether or not this is still doing what commonmark calls "loose" 🙃

It’s doing different things but it could be mapped to the same looseness definition!

> In the current approach say that a list node is treated as being loose if it or any of its list items is labeled with spread: true. If we stick with the label loose over spread, then we say that a list node is treated as being loose if it or any of its list items is labeled with loose: true.

Currently `loose` on a list is a combination of its own looseness and that of its children, potentially creating that inconsistent tree. With this change, it or it’s children could have spread, and its looseness could be calculated from that, always being consistent.

hessj1:
> @ikatyang Of course! I think we can do it as some tiny utility!

Is there a utility to allow for `loose` I am having issue after upgrading from v4 to 5@latest with my lists not being parsed into list items. like it used to.

PR #365, opened 2018-10-07
Replace remark-rewrite-header with remark-behead in the plugins list

strugee:
Note that I make this same recommendation downstream (i.e. remark-rewrite-header is deprecated).


wooorm:
Nice, thanks @strugee! 🎉

PR #367, opened 2018-10-24
remark-parse: turn block elements JSON into JS

rexxars:
We are using remark-parse for parsing markdown in react-markdown, and we've had a bunch of people report issues due to their (incorrect, I would say) webpack/rollup/current-bundler JSON configuration or handling.

Usually boils down to something like configuring json files to be treated as "raw text" and not scoping it to outside of node_modules. 

Since this default list of block elements isn't being generated (from what I can tell, anyway), I figured it might as well be JS? If you feel differently, feel free to reject this  - no hard feelings ;)


rexxars:
Many different issues. The last one was due to someone setting `/.json$/` to be loaded by `raw-loader` or something in their webpack config and not excluding `node_modules`.  ¯\_(ツ)_/¯

wooorm:
@rexxars Landed & released! Thanks!

PR #376, opened 2018-11-09
fix #374 & #375

### LINKED ISSUE #374: AtxHeading with trailing hashsigns is inconsistent with GFM
https://github.com/remarkjs/remark/blob/78b7aca650f041017b5dc5d1d72a7daa5376433a/packages/remark-parse/lib/tokenize/heading-atx.js#L107

GFM requires at least one space or tab between heading text and trailing hashsigns. For example:
```markdown
### awd###
### awd ###
```
will became

### awd ###

But in remark, the space is not required. The above two will be parsed into same ast:
```json
{
  "type": "root",
  "children": [
    {
      "type": "heading",
      "depth": 3,
      "children": [
        {
          "type": "text",
          "value": "awd"
        }
      ]
    },
    {
      "type": "heading",
      "depth": 3,
      "children": [
        {
          "type": "text",
          "value": "awd"
        }
      ]
    }
  ]
}
```

### LINKED ISSUE #375: Unescape '\\' is incorrect
https://github.com/remarkjs/remark/blob/78b7aca650f041017b5dc5d1d72a7daa5376433a/packages/remark-parse/lib/unescape.js#L29

I think it should be:
```js
index = value.indexOf(backslash, prev + 1)
```

For example:
````markdown
```\\\[awd
dwa
```
````

Remark will parse it into:
```json
{
    "type": "code",
    "lang": "[awd",
    "meta": null,
    "value": "dwa"
}
```

I think 'lang' should be `\[awd`.

TitanSnow:
Closes #374 
Closes #375 

There's a newly added test can't pass the stringify. It seems that there's a bug in remark-stringify, which I'm not familiar with.


wooorm:
@TitanSnow thank you so much! I split both PRs so that it’s more readable, and I think your heading fix could be more readable ([like so](https://github.com/remarkjs/remark/commit/c55e4a3b7023e0ceba16c8de3d74e48c94fba98a#diff-2eaedd32fca0ab69eef9a4d0722a5851R107)), but you’re credited as a co-author.

🥳

PR #381, opened 2019-01-10
Add remark-utf8s and remark-code-screenshot to list of plugins

Swizec:
Hi 👋

I just created two small plugins for Remark and thought they'd fit your list of existing plugins.


Swizec:
@sobolevn I copied that from the other lines and just verified with GitHub's online editor that it is exactly the same.

I don't know how else to fix it.

PR #382, opened 2019-01-11
Update remark-parse README: Location -> Position

mike-north:
In several places, a unist concept called "Location" is referred to, but the link is now broken, and the unist readme makes no mention of this type.

There is a [Position](https://github.com/syntax-tree/unist#position) type, which is probably what was intended. Perhaps the readme was written, and then "Location" was renamed to "Position" or something.


wooorm:
Yep, what you’re saying is exactly what happened! https://github.com/syntax-tree/unist/issues/13

PR #383, opened 2019-01-11
feat: TS type information for remark-stringify and remark-parse

mike-north:
This pull request adds typescript ambient type information for the `remark-stringify` package, as well as acceptance tests for types that's run in CI as a second Travis-CI stage

This change enables things TypeScript consumers to do things like this

```ts
import * as remarkStringify from 'remark-stringify'
/*
 * 👇 "ghost module" containing only type information, to avoid risk of colliding
 * w/ anything in a value-exporting module 
 */
import {RemarkStringifyOptions} from 'remark-stringify/types'
import * as unified from 'unified'

/** 
 * 👇 type-checking on the options object, including string literal types
 * for things like `fence`, `bullet`, etc...
 */
const stringifyOptions: Partial<RemarkStringifyOptions> = {
  gfm: true,
  bullet: '*',
  fences: true,
  fence: '`',
  incrementListMarker: false
}

unified().use(remarkStringify, stringifyOptions)
```


.travis.yml
```diff
@@ -1,15 +1,25 @@
 language: node_js
-node_js:
```
Rokt33r: I think updating travis.yml must be done in another pr. Those changes are irrelevant to typing.

packages/remark-stringify/package.json
```diff
@@ -23,6 +23,7 @@
     "index.js",
     "lib"
```
Rokt33r: Add type definition files to `files`.

packages/remark-stringify/package.json
```diff
@@ -40,14 +41,16 @@
...
     "tape": "^4.9.1",
     "unified": "^7.0.0",
     "unist-builder": "^1.0.3",
     "unist-util-visit": "^1.4.0",
     "wcwidth": "^1.0.1"
   },
   "scripts": {
-    "test": "tape test.js"
+    "test": "tape test.js",
+    "test:types": "dtslint types"
```
Rokt33r: `test:types` -> `test-types`

packages/remark-stringify/package.json
```diff
@@ -40,14 +41,16 @@
     "xtend": "^4.0.1"
   },
   "devDependencies": {
+    "dtslint": "^0.4.2",
```
Rokt33r: Install `typescript`

packages/remark-stringify/types/index.d.ts
```diff
@@ -0,0 +1,14 @@
+// TypeScript Version: 3.0
+
+declare module 'remark-stringify' {
```
Rokt33r: Use `declare namespace`

packages/remark-stringify/types/index.d.ts
```diff
@@ -0,0 +1,14 @@
...
+  import {RemarkStringifyOptions} from 'remark-stringify/types'
```
Rokt33r: Let's keep everything in one file.

packages/remark-stringify/types/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
Rokt33r: Use same options: https://github.com/vfile/vfile/pull/33/files#diff-c8a21e7e69c387b332b7871f4faef4da

packages/remark-stringify/types/tslint.json
```diff
@@ -0,0 +1,9 @@
+{
```
Rokt33r: Use https://github.com/vfile/vfile/pull/33/files#diff-65d6de72f3f0a20f873033fe5b908bb6

.travis.yml
mike-north: the types should be tested. Would you like one big command like

```
existing_tests && tests_for_types
```
instead of travis-ci jobs that run in parallel?

packages/remark-stringify/package.json
mike-north: Just curious, why? dtslint downloads and installs an array of TS versions just before tests?

packages/remark-stringify/types/index.d.ts
mike-north: Happy to co-locate everything to one source file if that's your preference.

Are you on board with using a "ghost module" as a safe way to expose type information (consistent with [DefinitelyTyped](http://definitelytyped.org/guides/best-practices.html) best practices). This would take the form of a single `.ts` file like

```ts
declare module 'remark-stringify/types' {
  ...
}

declare module 'remark-stringify' {
  import {RemarkStringifyOptions} from 'remark-stringify/types'

  ...
}

```

packages/remark-stringify/package.json
Rokt33r: I prefer to install it to control typescript version in our workspace(vs-code). Otherwise, we always have to use releases with `next` tag. And sometimes, it shows false alarm in my experience.(It might be fine for now because typescript become quite stable than two years ago. But I don't want to take the risk.)

If you know some reliable articles against my comment, it should be nice time to discard `typescript` from `devDependencies` of other repos too. 😉

packages/remark-stringify/types/index.d.ts
Rokt33r: It's a different issue. There are just two types to define types. One is using `declare module 'X'` and another one is using `declare namespace X`. The one I prefer is used by React and Angular. (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L48, https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/angular/index.d.ts#L33)  I just want to keep the consistency that we're using now.

And Babel won't read type definition files. 😄

packages/remark-stringify/types/index.d.ts
Rokt33r: The document looks quite old.

As long as I know the recommendation example,

```ts
declare module X
```

, now becomes

```ts
declare namespace X
```
> https://www.typescriptlang.org/docs/handbook/modules.html

Rokt33r: Let's keep `RemarkStringifyOptions` in `remark-stringify` rather than in  `remark-stringify/types`. I cannot understand why do you want to keep it in the different path.

mike-north: >  I cannot understand why do you want to keep it in the different path.

The benefit of a "ghost module" is you never have the potential to collide with current (or future) value exports from the `remark-stringify` module.

I'm happy to align with the maintainers' preferences, but if I were writing my own ambient types for my own library, I'd look for alignment with the standard DefinitelyTyped style

mike-north: I've tried this a couple of ways, and can't figure out how to get a commonJS style export (`export = foo`) along with a namespace that contains the options.

At this point, I'll have ask you to propose a concrete approach that you can verify works w/ the current (or similar) tests

packages/remark-stringify/package.json
mike-north: If you're looking for early warnings w.r.t. type breakage, `dtslint` is probably your best bet, since it tests the minimum supported version of the types and all newer versions (including @next).

For projects that use the TS compiler API, I'll typically set up a CI build matrix against multiple TS versions (i.e., https://github.com/code-to-json/code-to-json/blob/2b46e9fb685f8a71c2d906d81a7c389d9b108f8e/.travis/setup-ts-version.sh#L4-L6 ) since sometimes those [breaking changes](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes) are far more difficult to detect through conventional ts -> js usage.

If you want a local version of TS to allow vscode to work consistently, I'd suggest a separate PR with [a workspace config](https://github.com/code-to-json/code-to-json/blob/2b46e9fb685f8a71c2d906d81a7c389d9b108f8e/code-to-json.code-workspace#L3) that ensures all contributors benefit from this setup.

Rokt33r: > If you're looking for early warnings w.r.t. type breakage, dtslint is probably your best bet, since it tests the minimum supported version of the types and all newer versions (including @next).

I knew. But @next tagged releases has problems with vscode sometimes.

> If you want a local version of TS to allow vscode to work consistently, I'd suggest a separate PR with a workspace config that ensures all contributors benefit from this setup.

It doesn't fix my problem that coercing contributors to use same typescript version. 🙄 
We just make other users to use same typescript in their development environment for remark but not in their other projects. And installing it into `devDependencies` is the simplest solution as long as I know.

packages/remark-stringify/types/index.d.ts
Rokt33r: > The benefit of a "ghost module" is you never have the potential to collide with current (or future) value exports from the remark-stringify module.

It would be nice to have, but contributors also need to know where other interfaces exist.

> At this point, I'll have ask you to propose a concrete approach that you can verify works w/ the current (or similar) tests

 I'll make another pr to your pr. 👍

Rokt33r: Btw, thanks for your patient.

ChristianMurphy:
@mike-north @Rokt33r hows the discussion going on function props going?

mike-north:
Overhead of getting this merged is higher than I had hoped. These types work fine for me as a consumer of the library -- I'll have to re-engage later when time permits

Rokt33r:
@ChristianMurphy @wooorm I think we need to discuss which typescript version are we going to support. IMHO, it's kind of meaningless to support old version users for now status. Then, we could ignore my pr based on this pr.

wooorm:
@Rokt33r Over on unified (https://github.com/unifiedjs/unified/pull/53) you mentioned restarting work on this. How’s that going?

Rokt33r:
Sorry, I was busy...😢 I'll try to finalize this pr in 2 days!!

wooorm:
Perfect! Yes, remark is about to be updated with a major version, so that’s fine!

packages/remark-parse/types/index.d.ts
```diff
@@ -0,0 +1,52 @@
...
+  blockMethods: string[]
+  inlineTokenizers: {
+    [key: string]: remarkParse.Tokenizer
+  }
+  inlineMethods: string[]
+}
+
+declare namespace remarkParse {
+  interface Parse extends Attacher {
+    (options: RemarkParseOptions): void
```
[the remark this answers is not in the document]
mike-north: I have a test for the most common usage pattern here, which supports partial options
https://github.com/remarkjs/remark/pull/383/files#diff-e4f34330f5a7ed229a6b01775e638a1fR10. This is because a top type (`unknown`) is used here 

https://github.com/unifiedjs/unified/blob/38d2c8211e240a64891aea871caae376428d0200/types/index.d.ts#L21

Are there other usage patterns we aim to support? Is there a way to directly construct a `Parse` plugin, or is
```ts
unified().use(remarkParse, parseOptions)
```
the only way?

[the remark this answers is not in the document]
ChristianMurphy: > Unified types would highly benefit from a generic here. As written, options passed along with the use() call is a place where all type safety is lost.

I opened a PR with an initial implementation for what generic plugin settings could look like over at https://github.com/unifiedjs/unified/pull/58

Feel free to chime in or refactor.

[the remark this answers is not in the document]
Rokt33r: > Should this be Partial<RemarkParseOptions>?

Makes sense a lot. But rather than wrap it with `Partial`, I think we should make all props of the interface optional. @ChristianMurphy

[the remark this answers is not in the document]
ChristianMurphy: @mike-north unifiedjs/unified#58 has been merged and will be included in the next release.
Unified now has generics for plugins to declare their settings `interface`(s).

packages/remark-stringify/types/test.ts
```diff
@@ -0,0 +1,32 @@
+import remarkStringify = require('remark-stringify')
+import unified = require('unified')
+import {Node, Parent} from 'unist'
+
+const stringifyOptions: Partial<remarkStringify.RemarkStringifyOptions> = {
```
ChristianMurphy: with generic support, I don't think this needs to be explicitly cast, `unified().use(remarkStringify, stringifyOptions)` should verify.

packages/remark-parse/types/index.d.ts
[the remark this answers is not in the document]
ChristianMurphy: > Are there other usage patterns we aim to support? Is there a way to directly construct a Parse plugin, or is
> ```js
> unified().use(remarkParse, parseOptions)
> ```
> the only way?

As far as I know, `use` is the only way official to attach `remark-parse` to Unified.
@wooorm are there any others?

packages/remark-stringify/types/test.ts
Rokt33r: I'll fix it as soon as unified v8.1.0(with the modified types) is released.

wooorm: released!

ChristianMurphy: Updated

packages/remark-parse/types/index.d.ts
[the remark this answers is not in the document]
ChristianMurphy: Generic has been applied

packages/remark/types/test.ts
```diff
@@ -0,0 +1,13 @@
...
+
+interface PluginOptions {
+  example: boolean
+}
+
+const plugin = (options?: PluginOptions) => {}
+
+remark().use(plugin)
+remark().use(plugin, {example: true})
+remark().use({settings: {commonmark: true}})
```
ChristianMurphy: Looking this over, I wonder if Unified itself could have a generic so the settings can be validated. :thought_balloon: 


@wooorm is it accurate that presets has three forms?

```ts
// Three ways presets should be done
// plugins added via preset
unified().use({ plugins: [] })
// plugins and settings
unified().use({ plugins: [], settings: {} })
// parser and stringifier with settings
remark().use({ settings: {} })

//======
// This shouldn't happen
unified().use({ settings: {} })
```

packages/remark-stringify/types/test.ts
```diff
@@ -0,0 +1,50 @@
...
+  const original = visitors.heading
+
+  visitors.heading = heading
+
+  function heading(
+    this: unified.Processor,
+    node: Node & {depth: number},
+    parent?: Parent
+  ) {
+    return (node.depth === 2 ? '\n' : '') + original.apply(this, [node, parent])
```
ChristianMurphy: line 43 and 46 seem to indicate that plugins may have trouble accessing properties added by remark.
Using plain `Node` it gets the error.

```
ERROR: 46:13  strict-comparisons  Cannot compare type 'object' to type 'number
```

When extending `Node` it gets the error

```
Type '(this: Processor, node: Node & { depth: number; }, parent?: Parent | undefined) => string' is not assignable to type 'Visitor'.
  Types of parameters 'node' and 'node' are incompatible.
    Type 'Node' is not assignable to type 'Node & { depth: number; }'.
      Property 'depth' is missing in type 'Node' but required in type '{ depth: number; }'.
```

Thoughts on how different `Node` types from `remark`, `rehype`, `retext`, and `redot` could be supporter? :thought_balloon:

packages/remark/types/test.ts
ChristianMurphy: Opened https://github.com/unifiedjs/unified/pull/59 as one potential way to allow `remark` to declare settings typings for the parser and stringifier bundle.

packages/remark-stringify/types/test.ts
ChristianMurphy: One option would be a generic.
Something like

```ts
  interface Transformer<I extends Node = Node, O extends Node = Node> {
    (
      node: I,
      file: VFileCompatible,
      next?: (error: Error | null, tree: O, file: VFile) => {}
    ): Error | O | Promise<O>
  }
```

Which would guarantee that the input and the output are a super type of a unist `Node`.
But could not guarantee that type `I` matches what the previous plugin output, meaning it would be possible for JavaScript errors to slip through the typings.

---

Another would be to add a generic to Processor, and set that generic to a union of all types for the give language.
For example
```ts
type AllUnifiedTypes = Node | Parent
const unifiedProcessor = unified<Settings, AllUnifiedTypes>

type AllRemarkTypes = List | ListItem | Table | ...
const remarkProcessor = unified<Settings, AllRemarkTypes>
```

But then there would be the question of how do plugins participate in the supported node typings?
And how are transitions between languages, like mdast to hast, handled?

---

Other ideas?

Rokt33r: > Which would guarantee that the input and the output are a super type of a unist Node.
But could not guarantee that type I matches what the previous plugin output, meaning it would be possible for JavaScript errors to slip through the typings.

Looks awesome to me. 💯

Rokt33r: I don't think the another idea would work with plugin adding `AllUnifiedTypes`... The type can not be extended from plugins, unified or remark. So adopters should define their own `AllUnifiedTypes` based on plugins which they are using.

packages/remark/types/test.ts
wooorm:
```
unified().use({ settings: {} })
unified().use({})
```

...should not happen if remark-parse / remark-stringify (or some other plugin that uses `settings`) is not used. If a plugin defines that it uses settings somehow, these two cases would be OK (they both reset the settings).

packages/remark-stringify/types/test.ts
wooorm: @ChristianMurphy @Rokt33r What’s left to do for this PR to land?

ChristianMurphy: The main item is getting this test to pass.

The issue is currently the typing expects `visitor` to be passed a exactly a unist `Node`, when it may be a super type of `Node`, in this case a `HeadingNode` which also includes the `depth` property.

There is still some ongoing discussion into the approach.
The outcome of the discussion will likely also need to be applied to the `Transform` type in unified https://github.com/unifiedjs/unified/blob/e408747ec44fb541081e020a3842212e6a60f485/types/index.d.ts#L310-L316
Which similarly may be a super type of `Node`.

ChristianMurphy: If it sticks for too long, we have the option to put off adding visitor type.
Remove the test and that part of the typing, and add it in a follow up PR.

wooorm: Thabks for clarifying! The thread is becoming long so it's hard to know what's being discussed! I'd prefer to get *a* version out, so we can test this is the wild, and push the other breaking changes as well, and then make it perfect later. Could that work (without more breaking typescript changes soon?)

ChristianMurphy: Sure,
9011bf8891cf47c210509c4443baf1f38ff5b26f resolves the test issue, and adds a `FIXME` comment for what should be updated.
This is ready for a final review.

ChristianMurphy: Opened https://github.com/remarkjs/remark/issues/426 to continue the discussion on typing `visitors` marking this as resolved.

ChristianMurphy:
> Does remark itself need types?

@wooorm good call, remark itself also has types in the latest updates to this PR

PR #384, opened 2019-01-11
feat: TS type information for remark-parse 

mike-north:
NOTE: This branch is based on top of #383

This PR adds ambient type information and tests for the `remark-parse` package

Consumers of this library who benefit from TypeScript type information can do stuff like

```ts
import * as remarkParse from 'remark-parse'
import {RemarkParseOptions} from 'remark-parse/types'
import * as unified from 'unified'

/**
 * 👇 type-checking against the options object
 */
const parseOptions: Partial<RemarkParseOptions> = {
  gfm: true,
  pedantic: true
}

unified().use(remarkParse, parseOptions)

/**
 * 👇 type-checking against the options object
 */
const badParseOptions: Partial<RemarkParseOptions> = {
  gfm: 'true' // 🚨
}

```


Rokt33r:
I've merged this pr into https://github.com/remarkjs/remark/pull/383

PR #386, opened 2019-01-14
Update readme.md

revelt:
Let's supplement an example without pipes, so that people could clearly see how to turn AST into MD (string, not HTML), without chained functions. The current example is not very clear and I had to trawl unit tests to find this out.

### DIFF
--- a/packages/remark-stringify/readme.md
+++ b/packages/remark-stringify/readme.md
@@ -81,6 +81,26 @@ var processor = unified()
 process.stdin.pipe(createStream(processor)).pipe(process.stdout)
 ```
 
+Here is how you parse, assign the tree to a variable, log it, then render it back to markdown and log it again:
+
+```js
+import unified from "unified";
+import markdown from "remark-parse";
+import stringify from "remark-stringify";
+
+const tree = unified()
+  .use(markdown)
+  .parse("# Some title");
+
+console.log(`tree = ${JSON.stringify(tree, null, 4)}`);
+
+const doc = unified()
+  .use(stringify)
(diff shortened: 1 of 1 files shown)


packages/remark-stringify/readme.md
```diff
@@ -81,6 +81,26 @@ var processor = unified()
...
+  .use(markdown)
+  .parse("# Some title");
+
+console.log(`tree = ${JSON.stringify(tree, null, 4)}`);
+
+const doc = unified()
+  .use(stringify)
+  .stringify(tree);
+
+console.log(`doc = ${doc}`);
```
ChristianMurphy: Unified tends to encourage a plugin based approach.
To achieve the same effect with a plugin:

```js
import unified from "unified";
import parse from "remark-parse";
import stringify from "remark-stringify";

const doc = unified()
  .use(parse)
  .use(() => tree => console.log(`tree = ${JSON.stringify(tree, null, 4)}`))
  .use(stringify)
  .process("# Some title");

console.log(`doc = ${doc}`);
```
---

Having more recipes of how to handle common use cases, as well as some more complex patterns is a topic of ongoing discussion at https://github.com/syntax-tree/ideas/issues/7
Feel free to chime in with thoughts and comments.

wooorm:
@revelt Thanks for working on this. I do agree with Christian’s comment above.
All the example uses you found in the unit tests, are documented (and more) in unified’s docs, so instead of this PR I added more links to those.

revelt:
Awesome! Thank you @wooorm

PR #390, opened 2019-02-14
Clean leading tabs from new lines

### LINKED ISSUE #198: Sublists aren't parsed when indented with tabs
If you indent a sublist with `tabs`, they aren't parsed as lists.

Steps to reproduce:

```
git clone https://github.com/matheuss/remark-issue-198
cd remark-issue-198
npm i && node .
```

I'm happy to provide more info if needed ☺️

damccorm:
Right now sublists aren't parsed when indented with tabs. This change causes all leading tabs on new lines to be treated as 2 spaces.

Closes #198

### DIFF
--- a/packages/remark-parse/lib/parse.js
+++ b/packages/remark-parse/lib/parse.js
@@ -2,10 +2,13 @@
 
 var xtend = require('xtend')
 var removePosition = require('unist-util-remove-position')
+var vfile = require('vfile')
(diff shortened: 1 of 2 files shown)


damccorm:
Whoops, missed that this would cause other tests to fail outside of ```remark-parse``` 😳. I'll close here until I fix those and then try again.

PR #408, opened 2019-06-16
Add remark-tree-sitter to list of plugins

samlanning:
I've had a lot of fun the past couple of days creating a plugin that uses Tree-sitter to syntax-highlight code blocks. Feel like I've now got it to a state where i'm comfortable with other using it, and was hoping it could be added to this list! :)

https://github.com/samlanning/remark-tree-sitter

Feedback very welcome!


wooorm:
Wow that’s cool! ✨

/cc @ChristianMurphy check this out

Issue #415, opened 2019-07-09
Merge block quotes


arobase-che:
Hi,

As specified [gfm#220](https://github.github.com/gfm/#example-220).

When option `gfm` is used, we shouldn't merge two block quotes.
It's already done, if the option `commonmark` is used.

It's an easy thing to do, I will submit a PR soon.

❤️

wooorm:
Thanks for opening this, and that you’re writing a PR!

wooorm:
Closed by GH-417. ✨

PR #417, opened 2019-07-09
Merge blockquotes in gfm mode

arobase-che:
Hi,

The PR relative to #415.

Update some fixtures because gfm is by default.
The simplest fixtures is also updated to relate change ("in commonmark mode this block is not merged" become "in commonmark mode and gfm mode...").

That's all :D

### DIFF
--- a/packages/remark-parse/lib/tokenizer.js
+++ b/packages/remark-parse/lib/tokenizer.js
@@ -304,7 +304,7 @@ function mergeText(prev, node) {
 
 // Merge two blockquotes: `node` into `prev`, unless in CommonMark mode.
 function mergeBlockquote(prev, node) {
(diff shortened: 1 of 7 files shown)


packages/remark-parse/lib/tokenizer.js
```diff
@@ -304,7 +304,7 @@ function mergeText(prev, node) {
 
 // Merge two blockquotes: `node` into `prev`, unless in CommonMark mode.
```
fazouane-marouane: It looks like this comment needs to be updated

```suggestion
 // Merge two blockquotes: `node` into `prev`, unless in CommonMark or gfm modes.
```

arobase-che: Absolutely !

Issue #396, opened 2019-04-17
Fix support for single grave accent / backtick in table


zeke:
Hey @wooorm!

I think I just found a little bug in `remark` (or maybe `remark-html`)! A single backtick inside a table cell causes problems.

Consider this snippet:

```js
const input = `
| Shortcut | Description
|----------|------------
|<kbd>Ctrl</kbd><kbd>\`</kbd> | Open repository
`

const remark = require('remark')
const html = require('remark-html')

remark()
.use(html)
.process(input, function (err, output) {
  if (err) console.log(err)
  console.log(output)
})
```

For some reason, that backtick (`) is causing trouble for that table row. This is the output:

```html
<table>
  <thead>
    <tr>
     <th>Shortcut</th>
     <th>Description</th>
    </tr>
  </thead>
  <tbody>
    <tr>
        <td>
        <kbd>Ctrl</kbd>
        <kbd>`</kbd>
        | Open repository
        </td>
        <td></td>
    </tr>
  </tbody>
</table>
```

FYI this is not a showstopper bug for us. For now we can work around it in our project using the encoded form of the backtick: `&grave;`

zeke:
Thanks!

Issue #320, opened 2018-01-17
Table cell mis-parsing caused by empty `inlineCode`


ikatyang:
### Subject of the issue

```md
| col1 | col2 | col3 |
|---|--|--|
| long text | `` | text |
```

### Steps to reproduce

[astexplorer](http://astexplorer.net/#/gist/1d9f069160c787cfe8da508a76295f14/f7ae0fd331a570896ea408c443823855ef47ae5f)

### Expected behaviour

```json
{
  "type": "tableRow",
  "children": [
    {
      "type": "tableCell",
      "children": [
        {
          "type": "text",
          "value": "long text"
        }
      ]
    },
    {
      "type": "tableCell",
      "children": [
        {
          "type": "inlineCode",
          "value": ""
        }
      ]
    },
    {
      "type": "tableCell",
      "children": [
        {
          "type": "text",
          "value": "text"
        }
      ]
    }
  ]
}

```

### Actual behaviour

```json
{
  "type": "tableRow",
  "children": [
    {
      "type": "tableCell",
      "children": [
        {
          "type": "text",
          "value": "long text"
        }
      ]
    },
    {
      "type": "tableCell",
      "children": [
        {
          "type": "inlineCode",
          "value": ""
        },
        {
          "type": "text",
          "value": " | text"
        }
      ]
    }
  ]
}
```

wooorm:
GitHub renders it like this:

| col1 | col2 | col3 |
|---|--|--|
| long text | `` | text |

Which is *super* weird 🤔

wooorm:
Uhm, I think I get why GitHub is “weird”. Because there’s no good solution. Your expected behaviour is not going to work. I don’t like how GH does it, but we can change remark’s actual behaviour.

Take this example:

##### Markdown

```markdown
| 1 | 2 | 3 |
|-|-|-|
| a | `` | text |
| b | `` | `` |
| c | ` | ` |
| d | `|` | 
| e | `\|` | 
| f | \| | 
```

##### Rendered on GitHub

| 1 | 2 | 3 |
|-|-|-|
| a | `` | text |
| b | `` | `` |
| c | ` | ` |
| d | `|` | 
| e | `\|` | 
| f | \| | 

##### Things of note

* Inline code in tables is different because you can’t have vertical bars (`|`) inside inline code, in a table, the vertical bar always creates a new cell (except when escaped with a backslash, such as in row **e** and **f**)
[...]

Issue #295, opened 2017-09-06
remark add extra slash when there is `\` in table


wsdjeg:
here is the md file:

```markdown
| aa | bb | cc |
| -- | -- | -- |
| dd | ee / `\` | ff |
```

after run `remark --no-color test.md`

you will see

```markdown
| aa  | bb     | cc  |
| --- | ---------- | --- |
| dd  | ee / `\` \| ff |     |
```

but I think it should be 

```markdown
| aa  | bb       | cc  |
| --- | -------- | --- |
| dd  | ee / `\` | ff  |
```

PR #420, opened 2019-07-11
Fix inline code and vertical bars in tables

wooorm:
This PR fixes inline code in table cells. Previously. Inline code got preference over the table cell itself. This was added due to GH-73 in GH-83 four years ago.

That behaviour does not match current GFM and let to unexpected results in GH-295, GH-320, and GH-396.

Additionally, inline code cannot be empty in CM/GFM.

Related to GH-73.
Related to GH-83.
Closes GH-295.
Closes GH-320.
Closes GH-396.


wooorm:
/cc @ikatyang What do you think?

Also /cc @zeke: Do you think this may be a bug with how GH renders Markdown? Do you think empty code (see https://github.com/remarkjs/remark/issues/320#issuecomment-510634680) should be supported in cells? Should I raise this with GH / ping GFM maintainers?

ikatyang:
Since GFM is widely adopted, matching GFM should be better I think. And of course it's even better if it could be fixed on the GFM side.

PR #432, opened 2019-08-01
types: export remark options type

ChristianMurphy:
This was intended to be exported before but got missed.
It allows downstream adopters to reuse the full options type.


ChristianMurphy:
Example of potential use case https://github.com/electron/hubdown/pull/19#discussion_r309737195

PR #435, opened 2019-08-24
Fix line feeds in tables breaking them

wooorm:
Related-to syntax-tree/hast-util-to-mdast#56.
Related-to syntax-tree/hast-util-to-mdast#57.

### DIFF
--- a/packages/remark-stringify/lib/visitors/table-cell.js
+++ b/packages/remark-stringify/lib/visitors/table-cell.js
@@ -2,6 +2,10 @@
 
 module.exports = tableCell
 
+var lineFeed = /\n/g
+
 function tableCell(node) {
-  return this.all(node).join('')
+  return this.all(node)
+    .join('')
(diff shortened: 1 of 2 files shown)


packages/remark-stringify/lib/visitors/table-cell.js
```diff
@@ -2,6 +2,10 @@
 
 module.exports = tableCell
 
+var lineFeed = /\n/g
```
ChristianMurphy: Are carriage returns `\r` also a concern?

wooorm:
```suggestion
var lineFeed = /\r?\n/g
```

Right! Should be there as well.
`remark-parse` never puts first removes those, but also doesn’t put newlines in, so this edge case only happens if someone somehow injects them, and they could just as well inject carriage returns.

PR #440, opened 2019-10-02
Stop appending linefeeds on consecutive runs

oliveiraev:
If I run `remark-cli` several times on a file, each run will insert a final newline to the end of the checked target file.

This commit adds a verification to check if the file already ends with a linefeed and doesn't include additional ones when not needed.

### DIFF
--- a/packages/remark-stringify/lib/visitors/root.js
+++ b/packages/remark-stringify/lib/visitors/root.js
@@ -7,5 +7,9 @@ var lineFeed = '\n'
 // Stringify a root.
 // Adds a final newline to ensure valid POSIX files. */
 function root(node) {
-  return this.block(node) + lineFeed
+  var block = this.block(node)
+  if (block.substr(lineFeed.length * -1) === lineFeed) {
+    return block
+  }
+  return block + lineFeed
 }
--- a/packages/remark/test.js
+++ b/packages/remark/test.js
@@ -33,6 +33,14 @@ test('remark().processSync(value)', function(t) {
     'should accept stringify options'
   )
 
+  t.equal(
+    remark()
+      .processSync('<!-- last line\n')
+      .toString(),
+    '<!-- last line\n',
+    'should not add more than one linefeed at the end'
+  )
+
   t.throws(
     function() {
       remark()


wooorm:
This shouldn’t happen 🤔 What code are you using that keeps on adding an extra line feed?

oliveiraev:
> This shouldn’t happen  What code are you using that keeps on adding an extra line feed?

I use [VIM modelines](https://vim.fandom.com/wiki/Modeline_magic) at the end of my markdown files, and maybe the problem is that the comments are unclosed

https://asciinema.org/a/ec5mDFTHcvk09trCw8iKZFntL

wooorm:
Hmm, it is expected that the initial file, without an EOF EOL, has an EOL. It is also expected that after adding the comment, another EOF EOL is added. But the other EOLs are superfluous 🤔

oliveiraev:
> Hmm, it is expected that the initial file, without an EOF EOL, has an EOL. It is also expected that after adding the comment, another EOF EOL is added. But the other EOLs are superfluous

there **are** EOF EOL, they just doesn't show up. If they were missing, git diffs would report for the initial version...

```diff
diff --git a/README.md b/README.md
new file mode 100644
index 0000000..b45ef6f
--- /dev/null
+++ b/README.md
@@ -0,0 +1 @@
+Hello, World!
\ No newline at end of file
```

...and for the comment version

```diff
diff --git a/README.md b/README.md
index 8ab686e..706a8d7 100644
--- a/README.md
+++ b/README.md
@@ -1 +1,3 @@
 Hello, World!
+
+<!-- vim: set ai si sta et sw=4 sts=4 fenc=utf-8 nobomb eol ff=unix ft=markdown:
\ No newline at end of file
```

Also, after the very first run of `remark`, git would warn about changes due to the added EOF EOL.

oliveiraev:
I don't know typescript and can't make the build pass. Can you give me directions?

wooorm:
Ah, TS checking recently started failing: https://github.com/unifiedjs/unified/commit/9786410fd4355fc1821120f18f3e8d3b118daec3. I’ll work on a fix

wooorm:
The problem is the final HTML:

```javascript
var remark = require('./packages/remark')

var doc = `Hello, World!`
var index = -3

while (index++) {
  doc = remark()
    .processSync(doc)
    .toString()

  console.log([doc])
}
```

Yields:

```
$ node example.js 
[ 'Hello, World!\n' ]
[ 'Hello, World!\n' ]
[ 'Hello, World!\n' ]
```

wooorm:
Note btw, that VIM modelines, from the examples the doc you linked show, should be valid comments. `<!-- so you need the final stuff as well: -->`

wooorm:
I refactor the code a bit; thanks for working on this!

PR #441, opened 2019-10-12
Added remark-jargon plugin

joostdecock:
I wrote a plugin, so I added it to the list.
If that's not how things are supposed to work, please ignore this.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -110,6 +110,8 @@ See [Creating plugins][create] below.
     — add an improved image syntax
 *   [`remark-inline-links`](https://github.com/remarkjs/remark-inline-links)
     — change references and definitions to links and images
+*   [`remark-jargon`](https://github.com/freesewing/freesewing/tree/develop/packages/remark-jargon)
+    — inserts definitions for jargon terms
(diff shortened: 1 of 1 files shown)


wooorm:
Nice! I would like to suggest that you also include some more info on how remark users can install and use this, without gatsby!

P.S. `term.length` -> `term.length`!

joostdecock:
Yes, indeed. I need to setup a minimal repo to use this with remark only. It's not my use-case, so I haven't done it yet, but I have created an issue for it here: https://github.com/freesewing/freesewing/issues/163  I hope to get around to it soon.

Thanks for spotting the length typo.

joostdecock:
Ok, [README](https://github.com/freesewing/freesewing/tree/develop/packages/remark-jargon) updated. I've also setup a [minimal repo](https://github.com/joostdecock/remark-jargon-example) that shows its use.

I've fixed that typo too. I need to do a few other things before publishing an updated version but it will be done no later than tomorrow.

wooorm:
Nice! ✨

joostdecock:
Thank you 🙏

PR #451, opened 2019-11-09
Fix dead link to unified homepage

lf-:
The link to https://unified.js.org is dead and it appears to have moved to https://unifiedjs.com according to https://github.com/unifiedjs/unified.


wooorm:
Thanks @lf-!

PR #457, opened 2019-12-07
Add partial interface exports

### LINKED ISSUE #452: Improve types for option interfaces
### Improve types for option interfaces

Currently the interfaces for all options (e.g. [RemarkParseOptions](https://github.com/remarkjs/remark/blob/de3de6d3321da594406fd628e9452a237d04aafc/packages/remark-parse/types/index.d.ts#L23-L29)) require each setting to be defined.
As the options are populated by defaults (e.g. [index.js ](https://github.com/remarkjs/remark/blob/de3de6d3321da594406fd628e9452a237d04aafc/packages/remark-parse/index.js#L14)) not all settings are required to be defined.

Therefore the interfaces should look something like this:
```javascript
interface RemarkParseOptions {
  gfm?: boolean
  commonmark?: boolean
  footnotes?: boolean
  blocks?: string[]
  pedantic?: boolean
}

```

### Problem

I am using remark in a typescript setting, and these non-optional settings interface requires me to either re-implement the default options or use a not strictly typed generic dict.

### Expected behaviour

The settings on all option interfaces should be optional as they are not required to be set.
E.g. The following code segment should be a valid typescript code:
```typescript
parserSettings: RemarkParseOptions = {};
```

### Alternatives

As an alternative, it would be a possibility to export the default, so I can use the official defaults when implementing my options.

Roang-zero1:
With this an export for partial interfaces of option parameters is added.
These can be used in projects depending on remark to type their option parameters.
Fixes #452

### DIFF
--- a/packages/remark-parse/types/index.d.ts
+++ b/packages/remark-parse/types/index.d.ts
@@ -13,8 +13,8 @@ declare class RemarkParser implements Parser {
 }
 
 declare namespace remarkParse {
-  interface Parse extends Plugin<[Partial<RemarkParseOptions>?]> {
-    (options: Partial<RemarkParseOptions>): void
+  interface Parse extends Plugin<[PartialRemarkParseOptions?]> {
+    (options: PartialRemarkParseOptions): void
     Parser: typeof RemarkParser
   }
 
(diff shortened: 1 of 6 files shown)


wooorm:
What is the SemVer impact of this?

ChristianMurphy:
This leaves the existing interface and adds a new alias, I'd say SemVer Minor

PR #462, opened 2020-01-05
Add `remark-truncate-links` to list of plugins

CanRau:
Add [remark-truncate-links](https://github.com/GaiAma/Coding4GaiAma/tree/master/packages/remark-truncate-links)

shortens plain links in markdown not wrapped in `[]()`

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -195,6 +195,8 @@ See [Creating plugins][create] below.
     — highlight code blocks in Markdown files using
     [Tree-sitter](https://tree-sitter.github.io/tree-sitter/)
     (rehype compatible)
+*   [`remark-truncate-links`](https://github.com/GaiAma/Coding4GaiAma/tree/master/packages/remark-truncate-links)
+    — truncate/shorten urls not manually named
 *   [`remark-unlink`](https://github.com/remarkjs/remark-unlink)
     — remove all links, references, and definitions
 *   [`remark-unwrap-images`](https://github.com/remarkjs/remark-unwrap-images)


wooorm:
Nice! Two suggestions (also check out the rehype plugin suggestions).

- Add example input/output in the readme! I see three options but I don’t immediately see what they do?
- It depends on some nuance, but you can use [`mdast-util-to-string`](https://unifiedjs.com/explore/package/mdast-util-to-string/) to serialise `node` to text, so you won’t need to loop over all children!

CanRau:
Thanks for merging and feedback @wooorm really appreciated 🙏 

You mean replacing
```js
node.children?.map((child: NodeWithChildren): void => {
  if (typeof child?.value === `string` && child.value === node.url) {
    child.value = truncator(child.value, length, `...`)
  }
})
```
with something like
```js
const asString = toString(node)
if (asString === node.url) {
  node.children = [
    { type: 'text', value: truncator(asString, length, `...`) },
  ]
}
```
?

Actually I tried a few variations yet I don't think (yet) it can give me what I want as `mdast-util-to-string` will return `node.value` in this case then I wont be able to compare `node.url` & `child.value` to only truncate links where those match and therefore are not manually linked but automatically.
So I don't want to truncate
```
{
  type: 'link',
  title: 'https://github.com/GaiAma/Coding4GaiAma/tree/master/packages/gatsby-transformer-leasot#readme',
  url: 'https://github.com/GaiAma/Coding4GaiAma/tree/master/packages/gatsby-transformer-leasot#readme',
  children: [ { type: 'text', value: 'GitHub', position: [Position] } ],
  position: Position {
    start: { line: 133, column: 12, offset: 6958 },
    end: { line: 133, column: 115, offset: 7061 },
    indent: []
  }
}
```
because it's child value indicates that the markdown looks like `[GitHub](https..)`

Do you understand what I mean? But maybe I don't understanding what you're saying or how to do it 😅 

I'll push the new readme later today 🙌 

By the way congrats to the new-ish unifiedjs.com website just stumbled upon it yesterday, looks pretty cool 🎉 🚀 

Uh one other question, the plugin will currently set `node.title = node.title || node.url` is that okay or should that be a separate plugin to not enforce yet another transformation?

wooorm:
> “node.title = node.title || node.url”

I think another plugin makes more sense there? (also: browsers often do this already: users can hover over a link and the browser somewhere shows the URL)

> Congrats to the new-ish unifiedjs.com website

Thanks! Hope it’s useful! Let me know if you have feedback

> “I don't want to truncate`[GitHub](https..)`

Right! My proposed solution with `to-string` doesn’t do that I think?!

```ts
  return function(ast: NodeWithChildren): void {
    visit(ast, 'link', (node: NodeWithChildren) => {
      const value = toString(node)
      if (value === node.url) {
        node.children = [{type: 'text', value: truncator(value, length, `...`)}]
      }
    })
  }
```

...should do:
* `[GitHub](https..)` -> `[GitHub](https..)`
* `<https://example.com>` -> `[https://ex...](https://example.com)`
* `[https://example.com](https://example.com)` -> `[https://ex...](https://example.com)`

wooorm:
oh btw, another idea may be to use: https://github.com/sindresorhus/humanize-url

CanRau:
Not sure what I messed up, your code works like a charm 🎉 thanks 🙏 

All changes including the readme examples are now pushed & published 👍 

Uh yeah thanks for mentioning `humanize-url` looks pretty interesting, they're even planning to include truncation, I'll see when and how to integrate 👍 

Adding `tests` asap, too 🚀

wooorm:
Awesome!

CanRau:
> > “node.title = node.title || node.url”
> 
> I think another plugin makes more sense there? (also: browsers often do this already: users can hover over a link and the browser somewhere shows the URL)

Uh ya by the way, you're right of course about browsers generally showing the url 👍
From people around me most of them don't seem to know/recognize this though, so I thought when truncating it's just fair to still make it as easy as possible for people to verify the full URL before clicking.
That's obviously & sadly not really working on mobile like this.\
I sometimes long touch to verify a URL before deciding if I want to go there..

wooorm:
Ah, that's an interesting idea, haven't seen that, but maybe that's useful yeah!

CanRau:
You think I should put `node.title = node.title || node.url` back in then? Maybe having a switch to let people enable or disable it 🤔

wooorm:
I would personally do that in a different plugin, yeah!

PR #466, opened 2020-02-02
Add engines to package.json

MichaelDeBoey:
Also test against all possible versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
-  - lts/dubnium
+  - 8
+  - 10
+  - 12
   - node
 sudo: false
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 language: node_js
 node_js:
-  - 8
-  - 10
-  - 12
+  - lts/carbon
+  - lts/dubnium
+  - lts/erbium
   - node
-sudo: false
 deploy:
   provider: releases
   skip_cleanup: true
--- a/package.json
+++ b/package.json
@@ -88,5 +88,10 @@
       "./script/list-of-methods",
       "preset-wooorm"
     ]
+  },
+  "engines": {
+    "node": ">=8",
+    "npm": ">=6",
+    "yarn": ">=1"
   }
 }
--- a/packages/remark-cli/package.json
+++ b/packages/remark-cli/package.json
@@ -36,5 +36,10 @@
   "scripts": {
     "test": "tape test.js"
   },
-  "xo": false
+  "xo": false,
+  "engines": {
+    "node": ">=8",
+    "npm": ">=6",
+    "yarn": ">=1"
+  }
 }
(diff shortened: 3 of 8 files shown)


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
-  - lts/dubnium
+  - 8
+  - 10
+  - 12
```
ChristianMurphy: Testing across node versions is a welcome addition. :+1: 
Please use the LTS codenames `lts/carbon`, `lts/dubnium`, and `lts/erbium` over the number versions.
For consistency with other unified/remark projects.

package.json
```diff
@@ -88,5 +88,10 @@
       "./script/list-of-methods",
       "preset-wooorm"
     ]
+  },
+  "engines": {
+    "node": ">=8",
+    "npm": ">=6",
+    "yarn": ">=1"
```
ChristianMurphy: `unified` projects avoid setting the engines field, see https://github.com/unifiedjs/unified/issues/70#issuecomment-536666466 and https://github.com/unifiedjs/unified/issues/70#issuecomment-539406229 for details.

MichaelDeBoey: So you want these removed then? 🤔

wooorm:
Hi there, thanks for your first contribution to the unified collective (right?!) ✨

It’s generally a good idea to split branches and pull requests to do one thing only, now you’re doing three things:

- Updating a couple of things to use newer ES syntax
  I’m personally not interested in that, older syntax (callbacks, explicit object merging, etc), are fine.
  I’m of the opinion that old ES features are still very valid in a modern ES world, and typically more readable.
  They’re tests though, so it doesn’t matter much, but...
- Adding `engines` fields to `package.json`s
  Christian mentioned that already, and linked to reasoning why not to do it.
  We definitely can’t do this in a non-major: it’ll show errors for everyone that is on an older node or npm (there are millions of people using unified)
- Remove `sudo` from `.travis.yml`
  I believe this one is still needed, as this is an old project, and Travis changed the defaults on *newer* projects, I may be wrong tho but this was the case before.

So while I appreciate your work, I don’t think they make sense for this project at this time.

MichaelDeBoey:
@wooorm The newer ES syntax was due to `npm run test` failing when adding the `engines` fields.

I also removed `sudo` from `.travis.yml`, because it's deprecated

wooorm:
Thanks for the clarification and the work you put into this PR, appreciated.

Though, as stated [above](https://github.com/remarkjs/remark/pull/466#issuecomment-581363381), we’re not currently interested in adding the `engines` field restriction.

If you feel strongly about engines, I would be interested in hearing if there was some problem you ran into that could’ve been solved by having such a field? But for now, I’m closing this.

Thanks though!

MichaelDeBoey:
@wooorm I just wanted to add them to make it clearer what the minimum supported Node version is.
Since dropping support for a Node version should be a major release, I think we should just put all supported Node versions in the `engines` key.
If you decide to drop support for a certain version (because it's EOL or some other reason), you can bump that version and publish a major release.

Issue #470, opened 2020-02-21
Underscores within links interfering with emphasis in pedantic mode.


fabiosantoscode:
### remark seems to be conflating underscores in links with emphasis underscores, in pedantic mode.

When using pedantic mode, an underscore-surrounded link isn't properly parsed, when this link has underscores. Read below for repro, the code is clearer than my wording :)

### Your environment

*   **OS**: Elementary OS Luna (kernel 4.15.0-74-generic)
*   **Packages**: remark=10.0.1, remark-parse=6.0.3
*   **Env**: node=12.10.0, yarn=1.17.3

### Steps to reproduce

Run this in a node repl or script:

```
const Remark = require('remark')
new Remark().data('settings', {pedantic: true}).parse('_baz [link](https://example.com/with_underscore)_').children[0]
```

We can verify that the link is being parsed in a weird way, and the link is broken too, and starts pointing to `https://example.com/with`, which ignores its `_underscore` ending.

### Expected behaviour

This should return a simple tree, comprised of:

```
paragraph
  - emphasis
    - text "baz"
    - link [to the URL with an underscore]
      - text "link"
```

### Actual behaviour

I get a very complex tree:

```
paragraph
  - emphasis
    - text "baz"
    - linkReference
    - text "("
    - link to https://example.com/with
      - text "https://example.com/with"
  - text "underscore)_"
```

Cheers!

wooorm:
Hi! 👋 I would strongly suggest, if possible, to stop using pedantic mode: it emulates the buggy original markdown from 2004

fabiosantoscode:
@wooorm thank you for your suggestion. I will go through that avenue and check if the HTML is similar. Feel free to close!

wooorm:
If that's possible, that'd be great! It's been so long that I think it would be better for remark and markdown to start thinking of removing pedantic mode.
May I ask what the reason was for using this mode?

fabiosantoscode:
Gatsby [defaults it to true](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-remark/src/extend-node-type.js#L110). I don't know the reason behind it.

PR #477, opened 2020-03-27
Add deprecation note for `pedantic` mode

wooorm:
Closes GH-470.

pedantic mode is buggy. It won’t be in micromark. This PR warns users not to use it.

### DIFF
--- a/packages/remark-parse/readme.md
+++ b/packages/remark-parse/readme.md
@@ -175,20 +175,10 @@ referenced from inside other footnotes.
 
 ###### `options.pedantic`
 
-Pedantic mode (`boolean`, default: `false`).
-
-```markdown
-Check out some_file_name.txt
-```
-
-Turns on:
-
-*   Emphasis (`_alpha_`) and importance (`__bravo__`) with underscores in words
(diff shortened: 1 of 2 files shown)


wooorm:
I wanted to do a “soft” one first, just docs, to signal that the pedantic code is basically frozen for future issues or new users. Hoping the next major after this one will be micromark.

ChristianMurphy:
Agreed, code wise I was thinking something like Node's depercate
https://github.com/nodejs/node/blob/1b1c2993a5bc9a64cb4d417509d2fc35a64182f4/lib/internal/util.js#L62-L102
Where it is a soft deprecation, but adds some logs to signal the incoming change, in case people haven't reviewed the docs.

wooorm:
My plan was to be even softer mostly due to https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-remark/src/extend-node-type.js#L110 😬

ChristianMurphy:
ah, that make sense.
CC'ing the Gatby team members who have recently worked on the remark intergration to raise awareness.

/cc @pieh @sidharthachatterjee @wardpeet @DSchau

PR #484, opened 2020-03-30
Add `remark-slate` to list of plugins

hanford:


doc/plugins.md
```diff
@@ -201,6 +201,8 @@ See [Creating plugins][create] below.
     — custom syntax Wordpress- and Hugo-like shortcodes (new node type)
 *   [`remark-simple-plantuml`](https://github.com/akebifiky/remark-simple-plantuml)
     — convert PlantUML code to images
+*   [`remark-slate`](https://github.com/hanford/remark-slate)
+    — Serialize Markdown as a slate 0.50+ compatible object.
```
wooorm:
```suggestion
    — compile Markdown to Slate [nodes](https://docs.slatejs.org/concepts/02-nodes)
```

How about this?
- “Compile” because [unified@9](https://github.com/unifiedjs/unified/releases/tag/9.0.0) uses that term for the generic action, “serialize” is used for text
- Dropped the version number, as I’m assuming you’ll update the project when they change in the future?
- Added a link!

hanford:
Added the suggestions @wooorm - I'll change the serialize -> compile in the projects README.

Also want to say thanks for all of the work you've been doing, remark/unified are both really awesome projects.

wooorm:
Awesome! Thanks for making the plugin, and hope you'll enjoy unified!

Make sure to add 'remark-plugin' as a keyword btw, and as unified and remark both updated today it may be good to check everything is still working properly!

PR #485, opened 2020-04-07
Fixes lists with mixed indentation

### LINKED ISSUE #198: Sublists aren't parsed when indented with tabs

scinos:
Fixes #198

This PR introduces a new logic to determine the indentation stops when using tabs.

Before, a tab will only create an indentation stop for the whole tab. For example, a line starting with a tab character would have produced the stops ` {4: 0}`, meaning that the 4th indentation point (because tabSize=4) starts at index 0 (i.e. the first character). This was used to un-indent lists, by getting all lines that belong to the same `listItem` and looking for the biggest common indentation point. 

This is a problem when mixing spaces and tabs. Imagine a line indented with 2 spaces (it will have stops `{1:0, 2: 1}`, with 1st and 2nd indentation stops at characters 0 and 1, respectively. If the following line has an indentation of `{4: 0}` (using a tab), they won't have any common indentation stop. 

After this change, **we generate all intermediate indentation stops when using tabs**: A line starting with a tab will have stops `{1: 0, 2:0, 3:0, 4:0}`. This means that it the line above (or below) has an indentation of 1, 2, 3 or 4 spaces, both lines will be considered indented with the _same_ indentation.

The result is that, when "unindenting" a line starting with a tab, it can be "unindented" the equivalent to 1, 2, 3 or 4 spaces, and the tab will be consumed in the process.

### DIFF
--- a/packages/remark-parse/lib/util/get-indentation.js
+++ b/packages/remark-parse/lib/util/get-indentation.js
@@ -15,6 +15,7 @@ function indentation(value) {
   var character = value.charAt(index)
   var stops = {}
   var size
+  var lastIndent = 0
 
   while (character === tab || character === space) {
     size = character === tab ? tabSize : spaceSize
@@ -25,7 +26,10 @@ function indentation(value) {
       indent = Math.floor(indent / size) * size
     }
 
-    stops[indent] = index
+    while (lastIndent < indent) {
+      stops[++lastIndent] = index
+    }
+
     character = value.charAt(++index)
   }
 
--- a/packages/remark-parse/lib/util/remove-indentation.js
+++ b/packages/remark-parse/lib/util/remove-indentation.js
@@ -6,7 +6,6 @@ var getIndent = require('./get-indentation')
 
 module.exports = indentation
 
-var tab = '\t'
 var lineFeed = '\n'
 var space = ' '
 var exclamationMark = '!'
@@ -21,7 +20,6 @@ function indentation(value, maximum) {
   var index
   var indentation
   var stops
-  var padding
 
   values.unshift(repeat(space, maximum) + exclamationMark)
 
@@ -56,18 +54,7 @@ function indentation(value, maximum) {
         index--
       }
 
-      if (
-        trim(values[position]).length !== 0 &&
-        minIndent &&
-        index !== minIndent
-      ) {
-        padding = tab
-      } else {
-        padding = ''
-      }
-
-      values[position] =
-        padding + values[position].slice(index in stops ? stops[index] + 1 : 0)
+      values[position] = values[position].slice(stops[index] + 1)
     }
   }
 
--- a/test/fixtures/tree/mixed-indentation.commonmark.json
+++ b/test/fixtures/tree/mixed-indentation.commonmark.json
@@ -53,7 +53,7 @@
               "children": [
                 {
                   "type": "text",
-                  "value": "Very long\n\t\t\tparagraph",
+                  "value": "Very long\n\tparagraph",
                   "position": {
                     "start": {
                       "line": 3,
@@ -66,7 +66,7 @@
                       "offset": 48
                     },
                     "indent": [
-                      0
+                      2
                     ]
                   }
                 }
@@ -83,7 +83,7 @@
                   "offset": 48
                 },
                 "indent": [
-                  0
+                  2
                 ]
               }
             }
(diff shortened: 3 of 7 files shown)


packages/remark-parse/lib/util/get-indentation.js
```diff
@@ -25,7 +26,10 @@ function indentation(value) {
       indent = Math.floor(indent / size) * size
     }
 
-    stops[indent] = index
+    while (lastIndent < indent) {
```
scinos, line 29: For spaces (i.e. with `indent - lastIndent == 1`), this is equivalent to just `stops[indent]=index; lastIndent=indent`

For tabs, this will create entries in `stops` for all intermediate indents since the last one. For example, if the last indentation point is `1` (a space), and the next indentation point is `5` (a tab follows the initial space), this will add stops 2, 3, 4 and 5 pointing to character 1 (the tab).

packages/remark-parse/lib/util/remove-indentation.js
```diff
@@ -56,18 +54,7 @@ function indentation(value, maximum) {
...
-        index !== minIndent
-      ) {
-        padding = tab
-      } else {
-        padding = ''
-      }
-
-      values[position] =
-        padding + values[position].slice(index in stops ? stops[index] + 1 : 0)
+      values[position] = values[position].slice(stops[index] + 1)
```
scinos, line 57: This is not needed anymore because `stops` will now contain all indentation points, therefore `index` will never be different than `minIndent`.

test/fixtures/tree/mixed-indentation.commonmark.json
```diff
@@ -53,7 +53,7 @@
               "children": [
                 {
                   "type": "text",
-                  "value": "Very long\n\t\t\tparagraph",
+                  "value": "Very long\n\tparagraph",
```
scinos, line 56: This comes from trying to find the indentation of 
```
- Very long
<tab><tab>paragraph
```
This text is the "internal" content, the content without the indentation.

The resulting should be 
```
Very long
<tab>paragraph
```

as the `- ` in the first line marks the indentation, that is matched by the first `<tab>` in the second line.

Making this changes in the fixture makes it more consistent with the rest of the rest of the `mixed-indentation` fixtures, in all flavours and list styles.

scinos:
@transitive-bullshit I'd appreciate your feedback in this PR

thisconnect:
why is there a submodule tmp3 ?

Screen Shot 2020-04-07 at 8 06 45 PM

transitive-bullshit:
This looks really solid @scinos. Thank you so much for taking the time to work on this. I'm extremely booked with saasify stuff atm so I won't be able to give this the attention it deserves for awhile.

Glancing through, your approach and comments look reasonable and nothing jumps out at me as red flags. If all of the unit tests pass without regressions, that's the most important thing.

The only other note I have is that these changes should probably be accompanied by some additional unit tests that focus on this use case if there's anything not covered by existing tests.

This would fix https://github.com/sindresorhus/awesome-lint/issues/44 which has funding associated with it @sindresorhus 

@wooorm any additional thoughts?

wooorm:
@scinos The changes look good!

As only a couple of lines changed in the source code, and only a couple in the tests too, I do think adding new tests would be a great addition. For example, original markdown from https://github.com/Automattic/wp-calypso/pull/40798 may be useful, and then carefully comparing them with how [CM](https://spec.commonmark.org/dingus/) and [`markdown-it`](https://markdown-it.github.io) parse!

To add a test, add input markdown like `$name.text` in `input/`, make sure that `$name.json` is in `tree/` (you can also put `$name.commonmark.json` there to test CM too), and finally you can run `node script/regenerate-fixtures` to populate those fixtures! Then it’s often easier to compare your changes with `git diff`.

scinos:
Added a few tests for lists with mixed tabs and spaces.

I also run the tests in https://markdown-it.github.io/ and https://spec.commonmark.org/dingus/ , and compared them with printing the tree with `unist-util-inspect` (both with `commonmark:false` and `comomnmark:true`. The results are quite similar in all cases:

image

image

image

 Before my almost all lists in the new test failed to parse. Example:

image

scinos:
Any update on this? Is there anything else I should change in this PR?

wooorm:
Released as 8.0.1!

transitive-bullshit:
*Amazing* work @scinos!

sindresorhus:
@scinos Just saw this. Thanks! You got the IssueHunt reward: https://issuehunt.io/u/scinos

PR #495, opened 2020-05-18
PoC formatting is not validated in CI

remcohaszing:
This is a proof of concept to show formatting isn’t actually checked in CI.

A developer should run `prettier --write .` and `xo --fix` before committing, but CI should run `prettier --check .` and `xo` instead.

### DIFF
--- a/packages/remark/index.js
+++ b/packages/remark/index.js
@@ -1,6 +1,6 @@
 'use strict'
 
-var unified = require('unified')
+var unified = require(        'unified')
 var parse = require('remark-parse')
 var stringify = require('remark-stringify')
 


wooorm:
I believe you mentioned this in your spectrum post, and I may have responded there, but I forgot 😅 Could you expand on why you think it’s important to differentiate between the two? I get it a bit, but am wondering about your reasons.

I don’t think we’ve had many issues with it. And when it was an issue, it was often someone who didn’t really get node or npm scripts but still wanted to help, in which case  after a bit of back and forth and mentoring, I’m happy to do the work and run npm test before release, instead of having them do it.

BarryThePenguin:
There's probably a few different perspectives around code formatting? The two I can see here are..

1. Stylistic Issues - similar to the change in this pull request. The only problem it introduces is that it is inconsistent with existing code in the repository. These are easily fixed with `prettier --write .` and `xo --fix`, which can easily be run by maintainers after a PR is merged

2. Possible Errors - aren't usually fixed with `prettier --write .` and `xo --fix`, but they will be surfaced in pull requests by running `xo` in CI. These would need to be fixed by the developer before merging the pull request

I might have missed something, but I think the current setup covers both cases above?

Adding `prettier --check .` and `xo` to CI scripts would help uncover "Stylistic Issues" earlier in the pull request process, however they are easily fixed 👍 

---

One potential change I think we could make is, remove `prettier` and add [`prettier: true`](https://github.com/xojs/xo#prettier) to the `xo` configuration. This would also allow us to remove any `.prettierignore` files and additional configuration that we have

remcohaszing:
I didn’t consider that perspective. I actually thought this was a configuration error.

In my opinion all code should be properly formatted before being merged to master. However, I’m not a maintainer, so this decision isn’t up to me.

I would like to note that stylistic issues are not necessarily the same as auto fixable issues.

PR #502, opened 2020-06-09
TAWK-6522 Sub list fix

paultawk:

### DIFF
--- a/packages/remark-parse/lib/tokenize/list.js
+++ b/packages/remark-parse/lib/tokenize/list.js
@@ -157,8 +157,16 @@ function list(eat, value, silent) {
       index++
     }
 
-    if (item && size >= item.indent) {
-      indented = true
+    if (item) {
+      let requiredIndentCount = item.indent;
+
+      if (ordered) {
+        requiredIndentCount--;
+      }
+
+      if (size >= requiredIndentCount) {
+        indented = true
+      }
     }
 
     character = value.charAt(index)


ChristianMurphy:
Welcome @paultawk! :wave: 
Could you include some description of the issue you are facing and how this fixes it?
As well as a test case for the new functionality?

paultawk:
hey @ChristianMurphy 

i had some issue with how the markdown list and sublist is converted to html.

my setup is as follows

```javascript
// convert-to-html.js
'use strict';

const unified = require('unified');
const markdown = require('remark-parse');
const remark2rehype = require('remark-rehype');
const format = require('rehype-format');
const html = require('rehype-stringify');

const parser = unified();

parser.use(markdown);
parser.use(remark2rehype);
parser.use(format);
parser.use(html);

module.exports = function convertToHtml(content) {
  const node = parser.parse(content);
  const tree = parser.runSync(node);

  return parser.stringify(tree);
};
```

sample usage
```javascript
const convertToHtml = require('./convert-to-html');
const markdown = '1. Ordered list \n  * Sub list' // <- only 2 space indent

console.log(convertToHtml(markdown));

// result
/*
<ol>
  <li>Ordered list</li>
</ol>
<ul>
  <li>Sub list</li>
</ul>
*/

// desired result
/* 
<ol>
  <li>Ordered list
    <ul>
      <li>Sub list</li>
    </ul>
  </li>
</ol>
*/
```
https://github.com/remarkjs/remark/pull/502/files#diff-0b27a32611df78dd18ed7a0b170ec571L160
the result above would be 

```javascript
size = 2;
item.indent = 3
```

wooorm:
@paultawk I interpret what you’re saying as that the second list is currently adjacent to the first list (actual behavior), but you’d want the second list *inside* the first list (expected behavior), is that correct?

If so, it is CommonMark behavior and GFM behavior to treat them as adjacent lists (actual):

https://spec.commonmark.org/dingus/?text=1.%20Ordered%20list%20%0A%20%20*%20Sub%20list

To get the desired result, add an extra space before the `*`

wooorm:
> for reference the markdown comes from this editor http://demo.showdownjs.com/ with settings to vanilla

remark’s goal is to adhere to CommonMark. This behavior is documented there with [many examples](https://spec.commonmark.org/0.29/#example-264), and thus intentional. Showdownjs is an old project which hasn’t updated (this list) behavior since CM. We are slowly moving towards a world where more and more Markdown parsers are CM-compliant instead of having their own quirks; and that’s remark’s goal too.

Unfortunately, this PR would break away from that goal, so I can’t accept it.

paultawk:
@wooorm ok, thanks for the reply.

PR #513, opened 2020-07-18
Fix using remark in typescript directly

remcohaszing:
Remark exports a unified processor, not a function that returns a processor. Unified processors also happen to be functions returning a new processor, meaning the old type worked for most situations, but not all.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "tape": "^5.0.0",
     "tinyify": "^2.0.0",
     "typescript": "^3.0.0",
-    "unified": "^9.0.0",
+    "unified": "^9.1.0",
     "unist-builder": "^2.0.0",
     "unist-util-remove-position": "^3.0.0",
     "unist-util-visit": "^2.0.0",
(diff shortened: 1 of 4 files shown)


remcohaszing:
Perhaps we should create a new type in `unified`: `FrozenProcessor`.

`FrozenProcessor` would be what `Processor` is now, except the `use` method is omitted. `Processor` would inherit from `FrozenProcessor` and define `use`. `remark` would then export a `FrozenProcessor` instance.

ChristianMurphy:
That could work :+1: 

/cc @Rokt33r who implemented much of `Processor`, thoughts on having a `FrozenProcessor`?

remcohaszing:
Updated to use `FrozenProcessor`. I also had to rebase.

Rokt33r:
> /cc @Rokt33r who implemented much of Processor, thoughts on having a FrozenProcessor?

@ChristianMurphy I think it looks good for now. Sorry for my late answer

PR #514, opened 2020-07-18
Ignore package-lock.json files

remcohaszing:
They are generated by lerna, but the policy is to not commit them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ coverage
 /remark.js
 /remark.min.js
 yarn.lock
+package-lock.json


ChristianMurphy:
@remcohaszing did you find a case where lockfiles were generated?
The npmrc should be preventing that :thinking: 
https://github.com/remarkjs/remark/blob/da8203283a4759f29a58ea1782c16123f9805f0a/.npmrc#L1

BarryThePenguin:
I think running `npm install` from a package directory would generate one

wooorm:
Yeah, I would think `.npmrc` should’ve prevented it indeed, but as @remcohaszing saw this pop up I’m assuming this was somehow not working 100% 🤷‍♂️

remcohaszing:
There’s a top level `.npmrc` file. Since the subpackages are installed using `lerna`, their `package-lock.json` files were still being generated.

PR #515, opened 2020-07-18
Fix to make all remark options optional

remcohaszing:
This matches the their respective call signatures.

The `Partial*Options` variants are now deprecated aliases to their matching types for backwards compatibility.


ChristianMurphy:
@remcohaszing when you have a chance could you looking into the CI error?

wooorm:
Thanks, going to release when the other type PRs land!

PR #539, opened 2020-10-13
Update types for remark@next

wooorm:
Here are the type updates for the new remark-parse, remark-stringify, remark!

I tried to leave the type tests as much as they were, to reduce noise in the diff—we can clean them later.

### DIFF
--- a/packages/remark-parse/types/index.d.ts
+++ b/packages/remark-parse/types/index.d.ts
@@ -1,11 +1,11 @@
 // TypeScript Version: 3.0
 
-import {Parser, Plugin} from 'unified'
+import {ParserFunction, Plugin} from 'unified'
 import {Options} from 'mdast-util-from-markdown'
 
 declare namespace remarkParse {
   interface Parse extends Plugin<[RemarkParseOptions?]> {
-    Parser: Parser
+    Parser: ParserFunction
   }
 
   type RemarkParseOptions = Options
--- a/packages/remark-parse/types/index.d.ts
+++ b/packages/remark-parse/types/index.d.ts
@@ -1,12 +1,10 @@
 // TypeScript Version: 3.0
 
-import {ParserFunction, Plugin} from 'unified'
+import {Plugin} from 'unified'
 import {Options} from 'mdast-util-from-markdown'
 
 declare namespace remarkParse {
-  interface Parse extends Plugin<[RemarkParseOptions?]> {
-    Parser: ParserFunction
-  }
+  interface Parse extends Plugin<[RemarkParseOptions?]> {}
 
   type RemarkParseOptions = Options
 }
--- a/packages/remark-parse/types/index.d.ts
+++ b/packages/remark-parse/types/index.d.ts
@@ -1,113 +1,16 @@
 // TypeScript Version: 3.0
 
-import {Node, Parent, Position} from 'unist'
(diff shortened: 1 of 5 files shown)


ChristianMurphy (approved):

packages/remark-parse/types/index.d.ts
```diff
@@ -1,113 +1,16 @@
...
-  /**
-   * @deprecated Use `RemarkParseOptions` instead.
-   */
-  type PartialRemarkParseOptions = RemarkParseOptions
-
-  interface Add {
-    (node: Node, parent?: Parent): Node
-    test(): Position
-    reset(node: Node, parent?: Node): Node
+    Parser: Parser
```
ChristianMurphy, line 8: double checking, the interface for parser is unchanged?

packages/remark-stringify/types/index.d.ts
```diff
@@ -1,51 +1,14 @@
...
-  }
-}
+import {Compiler, Plugin} from 'unified'
+import {Options} from 'mdast-util-to-markdown'
 
 declare namespace remarkStringify {
   interface Stringify extends Plugin<[RemarkStringifyOptions?]> {
-    Compiler: typeof RemarkCompiler
-    (this: Processor, options?: RemarkStringifyOptions): void
+    Compiler: Compiler
```
ChristianMurphy, line 8: double checking, the interface for compiler is unchanged?

wooorm: They used to be classes, now they’re the simpler function version. Similar to rehype. I was under the impression this would be the correct way to type that?

wooorm: I guess I could import `ParserFunction` / `CompilerFunction`?

ChristianMurphy: Does it still support classes? Or is it only the function version now?
If it's only function, I'd lean towards pulling in CompilerFunction directly, otherwise Compiler is good.

wooorm: No classes I guess?

These two should be something for unified to interface with though, not really for other users 🤔

Do you have an example of how this would look? rehype-{parse,stringify} also use the function-version, but don‘t use function types: https://github.com/rehypejs/rehype/blob/main/packages/rehype-stringify/types/index.d.ts

wooorm: > If so why is it on the public interface?

No clue! I guess it was already exposed, and so are the rehype ones 🤷‍♂️

wooorm: I pushed to use function versions, but it seems like we could just as well drop them, because there is no such field that can be used?

```
  interface Parse extends Plugin<[RemarkParseOptions?]> {
_    Parser: ParserFunction
  }
```

ChristianMurphy (approved):

packages/remark-stringify/types/index.d.ts
ChristianMurphy: I think we could do that, I think Parser and Compiler used to exist to allow extending the parser.
Since that is no longer the extension point, I think it can be safely be removed :crossed_fingers:

wooorm:
This is now released in [`remark@13.0.0`](https://github.com/remarkjs/remark/releases/tag/13.0.0)

PR #553, opened 2020-10-26
Add `remark-extract-toc` and `remark-slate-transformer` to list of plugins

inokawa:
I made 2 plugins for remark. I checked that it works with remark with micromark.
Thank you for this great library and ecosystem.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -94,8 +94,6 @@ The list of plugins:
     — add `target` and `rel` attributes to external links
 *   🟢 [`remark-extract-frontmatter`](https://github.com/mrzmmr/remark-extract-frontmatter)
     — store front matter in vfiles
-*   🟢 [`remark-extract-toc`](https://github.com/inokawa/remark-extract-toc)
-    — store table of contents
 *   🔴 [`remark-fenced-divs`](https://github.com/benabel/remark-fenced-divs)
     — new syntax for Pandoc’s `fenced_divs` support (new node type, rehype
     compatible)
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -208,7 +208,8 @@ The list of plugins:
 *   🟢 [`remark-slate`](https://github.com/hanford/remark-slate)
     — compile Markdown to [Slate nodes](https://docs.slatejs.org/concepts/02-nodes)
 *   🟢 [`remark-slate-transformer`](https://github.com/inokawa/remark-slate-transformer)
-    — compile Markdown to [Slate nodes](https://docs.slatejs.org/concepts/02-nodes), and also Slate nodes to Markdown
+    — compile markdown to [Slate nodes](https://docs.slatejs.org/concepts/02-nodes)
+    and Slate nodes to markdown
 *   🟢 [`remark-slug`](https://github.com/remarkjs/remark-slug)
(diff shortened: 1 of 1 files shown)


wooorm:
Cool! Thanks for your kind words. And congrats on your first plugins!

#### `remark-extract-toc`

- `commonmark: true`, which is in the example, no longer does anything. It was removed in the [recent major](https://github.com/remarkjs/remark/releases/tag/13.0.0).
- I suggest using the variable name `extractToc` instead of `toc`, because folks otherwise might think it’s `remark-toc`
- Have you thought about using [`mdast-util-toc`](https://github.com/syntax-tree/mdast-util-toc)? It has all the features to generate a toc too, which you

Importantly, it seems this “plugin” will not work for a whole `process`, because it doesn‘t return a node (there are no types). We [discourage using the `remark-` prefix in that case, because it could confuse users](https://github.com/remarkjs/remark/blob/main/doc/plugins.md#creating-plugins).

Could you tell me more about why you made this project, so that we can find a way for this to be a proper plugin, and added to this list?

#### `remark-slate-transformer`

This project looks great!! ✨

inokawa:
@wooorm 

> Could you tell me more about why you made this project, so that we can find a way for this to be a proper plugin, and added to this list?

What I wanted to do is to create nested tree data (including only headings) from a markdown document, and render it with something like React.
It looks similar to what in this page on the right pane. -> https://formik.org/docs/overview
I couldn't find out the way so I made it, but I don't know it is common usecase...

> - commonmark: true, which is in the example, no longer does anything. It was removed in the recent major.
> - I suggest using the variable name extractToc instead of toc, because folks otherwise might think it’s remark-toc

Thank you for the suggestion. I fixed them!

wooorm:
> I couldn't find out the way so I made it, but I don't know it is common usecase...

I’ve hear about it before, yeah!

In that case, I’d suggest to return just the toc (from mdast-util-toc, or yourself) but as valid mdast. That could then be turned into hast, or whatever, and follow the typical unified flow!

inokawa:
> In that case, I’d suggest to return just the toc (from mdast-util-toc, or yourself) but as valid mdast. That could then be turned into hast, or whatever, and follow the typical unified flow!

So if the output is something like nested List with ListItem, or Paragraphs not nested, would be better as a plugin.
Anyway not suitable at least now, so I remove it from the list. Thank you.

doc/plugins.md
```diff
@@ -207,6 +207,8 @@ The list of plugins:
     — turn PlantUML code blocks to images
 *   🟢 [`remark-slate`](https://github.com/hanford/remark-slate)
     — compile Markdown to [Slate nodes](https://docs.slatejs.org/concepts/02-nodes)
+*   🟢 [`remark-slate-transformer`](https://github.com/inokawa/remark-slate-transformer)
+    — compile Markdown to [Slate nodes](https://docs.slatejs.org/concepts/02-nodes), and also Slate nodes to Markdown
```
wooorm, line 211:
```suggestion
    — compile markdown to [Slate nodes](https://docs.slatejs.org/concepts/02-nodes)
    and Slate nodes to markdown
```

wooorm:
Nice, thanks @inokawa!

And let me know if you have further Qs on `remark-extract-toc`!

PR #555, opened 2020-10-26
Update list of plugins emoji legend for deuteranopia

kevin940726:
As a person who has minor color-blind disability, I can't easily distinguish between the red circle (🔴 ) with the green circle (🟢 ). I figured it could be easier for people like me to have them replaced with other emojis. The changes are as follow:

- 🟢  -> ✅ 
- 🟡  -> ⚠️ 
- 🔴  -> 🔴 (unchanged)

I think it's a better representation to avoid the color only changes. Thoughts :)?

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -26,7 +26,8 @@ Some plugins are affected by the recent switch in the underlying parser of
 remark.
 Their status is encoded below as:
 
-*   🟢 This plugin **is not affected** and works the same as before, or this plugin is affected: **a new version is already released**
+*   🟢 This plugin **is not affected** and works the same as before,
+or this plugin is affected: **a new version is already released**
 *   ⚠️ This plugin is affected: it’s **currently broken** and maintainers
     have been notified
 
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -26,241 +26,240 @@ Some plugins are affected by the recent switch in the underlying parser of
 remark.
 Their status is encoded below as:
 
-*   ✅ This plugin **is not affected** and works the same as before
-*   ⚠️ This plugin is affected: **a new version is already released**
-*   🔴 This plugin is affected: it’s **currently broken** and maintainers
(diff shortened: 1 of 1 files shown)


wooorm:
Nice, good idea! The test failure is unrelated.

I’m not too happy with using `⚠️` for something that is already fixed.

Could we instead use `🟢` for both plugins that are not affected, and plugins that are fixed already, and then `⚠️` for `🔴` (This plugin is affected: it’s **currently broken** and maintainers have been notified)?

kevin940726:
> Could we instead use 🟢 for both plugins that are not affected, and plugins that are fixed already, and then ⚠️ for 🔴 (This plugin is affected: it’s currently broken and maintainers have been notified)?

Does that mean to group _not-affected_ and _already-fixed_ plugins together in the same category (🟢 )? I'm okay with that! Will push a commit soon.

wooorm:
Yes, that's what I meant! Thanks!

kevin940726:
Feel free to adjust the wording of the legends to make it clearer :).

doc/plugins.md
```diff
@@ -26,29 +26,29 @@ Some plugins are affected by the recent switch in the underlying parser of
 remark.
 Their status is encoded below as:
 
-*   🟢 This plugin **is not affected** and works the same as before
-*   🟡 This plugin is affected: **a new version is already released**
-*   🔴 This plugin is affected: it’s **currently broken** and maintainers
+*   🟢 This plugin **is not affected** and works the same as before,
+or this plugin is affected: **a new version is already released**
```
wooorm, line 30:
```suggestion
*   🟢 This plugin **was not affected** or **a new version is already released**
```

wooorm (approved):

wooorm:
[This is great](https://github.com/remarkjs/remark/blob/main/doc/plugins.md#list-of-plugins), much more readable too!

PR #706, opened 2021-05-05
Add `remark-img-links` to list of plugins

Pondorasti:
## Update Documentation

I've recently written a new `remark` plugin,[`remark-img-links`](https://github.com/Pondorasti/remark-img-links), that I needed for a personal project. I thought it might be a good idea to add it to the "list of plugins" [document](https://github.com/remarkjs/remark/blob/main/doc/plugins.md).

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -137,6 +137,8 @@ The list of plugins:
     — new syntax to create iframes (new node type, rehype compatible)
 *   🟢 [`remark-images`](https://github.com/remarkjs/remark-images)
     — add an improved image syntax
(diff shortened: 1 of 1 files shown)


wooorm (approved):

wooorm:
Nice!

Might it be an idea to:
- Make sure that full absolute URLs don’t break? `https://example.com` turns into `https://mywebsite.comhttps://example.com`
- For proper URL resolving, you can use [`new URL(thing, base).href`](https://nodejs.org/api/url.html#url_new_url_input_base), which might be a bit better than concating strings

Pondorasti:
Thanks for the suggestion, already made the change!

PR #826, opened 2021-08-27
Add `remark-torchlight` to list of plugins

aarondfrancis:
### Initial checklist

### Description of changes

Adds [`remark-torchlight`](https://github.com/torchlight-api/remark-torchlight) to plugins list. Thanks!

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -245,6 +245,8 @@ The list of plugins:
     — check and add the document title
 *   🟢 [`remark-toc`](https://github.com/remarkjs/remark-toc)
     — add a table of contents
+*   🟢 [`remark-torchlight`](https://github.com/torchlight-api/remark-torchlight)
+    — syntax highlighting powered by [torchlight.dev](https://torchlight.dev)
 *   🟢 [`remark-tree-sitter`](https://github.com/samlanning/remark-tree-sitter)
     — highlight code blocks in markdown files using
     [Tree-sitter](https://tree-sitter.github.io/tree-sitter/)


wooorm:
I’m fine-ish with adding this, but perhaps it’s better as a rehype plugin?
E.g., we’ve marked [`remark-highlight.js` as legacy](https://github.com/remarkjs/remark-highlight.js).
I think a `rehype-torchlight` plugin might be better?

aarondfrancis:
Honestly, I have a hard time understanding the whole remark, rehype, unified ecosystem, so I can't really offer an opinion 😬 

If you can explain in a bit more detail what it is you're proposing, I can give it a shot. I'm traditionally a backend developer and trying to navigate the JS ecosystem has my head spinning a bit, but I'm doing my best.

Let me know how I can help

aarondfrancis:
I was trying to get it to work with this example: https://github.com/vercel/next.js/blob/canary/examples/blog-starter/lib/markdownToHtml.js

wooorm:
remark is for markdown. If you need to handle markdown specifics, such as the two different link types (`[a](url)` and `[a][b]\n\n[b]: url`), remark/mdast/markdown is for that.
rehype is for HTML. If you care about `a` being a link, that’s where rehype/hast/html are good for.

With an analogy: say you have code in TypeScript. You want to replace an `isProduction` variable with `true` or `false` depending on your build. Now, you could solve this with a plugin that runs on TS code, which has access to type annotations and such. Or you can run it on JS code.
For this example, it doesn’t really make sense in TS. It’s probably easier in JS land, as you don‘t have to handle type annotations, and you can share it with more people: more people use JS (at some stage), than TS.

This is similar. Perhaps someone is writing HTML and wants to process that. There’s also [orga](https://github.com/orgapp/orgajs) for org mode already that is similar to remark, but a different format, and also uses rehype plugins. That community could then also use your plugin!

aarondfrancis:
Thanks for the explanation. I guess I can make a new plugin for those places, but would you mind adding the one I've already done here?

Also I no idea what "org mode" is. Just another ecosystem to learn! 😅

wooorm:
> Just another ecosystem to learn! 

It’s not specifically from *us*, but it’s compatible. It’s a different language that several people prefer over markdown

wooorm (approved):

wooorm:
Note that I might at some point stop recommending projects like this in remark space. They keep on perpetuating this complexity of where things should go for both users and potential makers!

PR #933, opened 2022-01-13
Add `remark-docx` and `remark-pdf` to list of plugins

inokawa:
### Initial checklist

### Description of changes

Add `remark-docx` and `remark-pdf` to list of plugins

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -93,6 +93,8 @@ The list of plugins:
     — new syntax for directives (generic extensions)
 *   🟢 [`remark-directive-rehype`](https://github.com/IGassmann/remark-directive-rehype)
     — turn [directives][d] into HTML custom elements (rehype compatible)
+*   🟢 [`remark-docx`](https://github.com/inokawa/remark-docx)
+    — compile markdown to docx
 *   🟢 [`remark-dropcap`](https://github.com/brev/remark-dropcap)
     — fancy and accessible drop caps
 *   🟢 [`remark-embed-images`](https://github.com/remarkjs/remark-embed-images)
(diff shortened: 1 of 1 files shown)


wooorm (approved):
Nice, thanks! Cool plugins.

For docx:
- I believe Node.js just added support for Blobs, maybe you can describe that somewhere, if that’s also possible now?
- wouldn’t there be more options on how to turn markdown into docx?
- Does it make sense to instead do this in rehype? Turn HTML into Docx?
- It’s best to *either* transform, *or* compile, instead of doing both. I understand this case though (`imageResolver`). Maybe it should be a separate plugin?

For pdf:
- same about buffers and blobs, and about options, and about HTML perhaps?

inokawa:
> I believe Node.js just added support for Blobs, maybe you can describe that somewhere, if that’s also possible now?

Exactly Blob is new in Node.js so it's seems better to explain it. Thank you!

> wouldn’t there be more options on how to turn markdown into docx?
> Does it make sense to instead do this in rehype? Turn HTML into Docx?

Markdown / HTML and docx / pdf has very very different document structure.
I don't know how hard in HTML, but in my opinion, conversion from markdown is easier because markdown has fewer syntaxes than HTML and do not have CSS, which is difficult to know extact style without rendering on browser.
So markdown does not include style definition and I'm thinking to add options to style page / header / paragraph / etc.

And I think one of the popular way to generate pdf is HTML and puppeteer.
However skia in chromium cannot handle some fonts like Japanese well so it would be OK to seek another way (currently remark-pdf cannot handle Japanese though, because it needs custom font and the pdf generation library used in this plugin is bit tricky...).
https://bugs.chromium.org/p/skia/issues/detail?id=244

> It’s best to either transform, or compile, instead of doing both. I understand this case though (imageResolver). Maybe it should be a separate plugin?

Something like what replace image urls with ArrayBuffer?
It may be possible but I'm not sure yet.

PR #982, opened 2022-04-16
Adds remark-mdx-postcss plugin to list

SSHari:
### Initial checklist

### Description of changes

Add the `remark-mdx-postcss` plugin to the plugins page. The plugin is designed to create a dynamic style tag for MDX which makes it easier to create isolated MDX bundles.

[remark-mdx-postcss](https://github.com/SSHari/remark-mdx-postcss)

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -169,6 +169,8 @@ The list of plugins:
     — new syntax for math (new node types, rehype compatible)
 *   🟢 [`remark-mdx`](https://github.com/mdx-js/mdx/tree/main/packages/remark-mdx)
     — support MDX (JSX, expressions, ESM)
+*   🟢 [`remark-mdx-postcss`](https://github.com/SSHari/remark-mdx-postcss)
+    — create a dynamic style tag for MDX via postcss
 *   🟢 [`remark-mermaidjs`](https://github.com/remcohaszing/remark-mermaidjs)
     — transform mermaid code blocks into inline SVGs
 *   🟢 [`remark-message-control`](https://github.com/remarkjs/remark-message-control)


wooorm:
Hi there!

1.  I believe The types are correct [here](https://github.com/SSHari/remark-mdx-postcss/blob/9fbc428a6e5774768deb3384fdd8cebe3b6d58a0/src/index.ts#L20), `mdxJsxFlowElement` is supposed to have different children: headings, paragraphs, etcetera. It’s not supposed to have a `text` as a direct child.
2.  I don’t think a `text` node is the right way to inject CSS into a JSX element. To clarify, you are creating an AST for:
    ```jsx
    <style>
      the css
    </style>
    ```
    
    but you might know that that would fail depending on what’s included in the CSS. A better way to include any string would be:

    ```jsx
    <style>{`the css`}</style>
    ```
3. I am wondering whether `remark` is the correct place to do this, as opposed to rehype?
   As this plugin injects a JSX element for an HTML element in markdown, but we have a whole HTML ecosystem one step later, the problem this plugin solves could be easier and clearer to solve as a rehype plugin, with the added benefit of making the plugin useable for more people (as there are no MDX specific features).
4. With a name of `remark-mdx-postcss`, and other users finding this, but this plugin not doing MDX-specific things yet, I wonder what it could do? Would people want to include `<style>` elements in their MDX? Would they want to export CSS as a string in a file? Are there other tailwind or React (which is frequently but not always used with MDX) things that could be useful?

SSHari:
Hey @wooorm!

Thanks for the feedback. I think it's probably worthwhile to provide you a little background around the problem I was trying to solve. In my case, I have MDX content which I store remotely in a database. I pull this content on demand and bundle it for a React application using a tool like [mdx-bundler](https://github.com/kentcdodds/mdx-bundler). This MDX often contains tailwind styles defined directly in the content. The problem is that since I don't know the styles ahead of time, I have to generate styles on demand for the MDX content which is what this plugin was designed to do. Here's an example of how I'm [generating the tailwind styles](https://github.com/SSHari/remark-mdx-postcss/tree/9fbc428a6e5774768deb3384fdd8cebe3b6d58a0#example-build-tailwind-styles-dynamically).

Now with that said, I think you're right about this being better as a `rehype` plugin. I'm sure I got a bit ahead of myself since my use case is driven by content that starts as MDX, but you're correct that all I'm really needing to do here is use postcss to generate a style tag that I can include with my final output which isn't MDX specific.

I can close this PR and go down the path of implementing this as a `rehype` plugin instead, but before I do that I was hoping you might be able to provide some thoughts on a few things so that I'm better prepared to create the plugin.

1. Taking a look at the available node types for `rehype`, it seems like for the most part the logic for this can be the same. The primary difference would be the node that would get created. Perhaps something like the following:
```js
function rehypePostCssStyleTag() {
	return transformer(tree, file) {
		const css = // postcss processing
		tree.children.unshift({
			type: 'element',
			tagName: 'style',
			children: [{ type: 'text', value: css }]
		});
	}
}
```

2. My use case is to use postcss to generate tailwind styles that are appended via a style tag. I originally thought it might be good to generalize and allow any postcss plugins, but maybe that's not a great idea. Do you think there's any issue with me trying to generalize here or do you recommend that these plugins try to handle specific use cases?

Thanks for all the help 🙂!

wooorm:
Hey again :)

> This MDX often contains tailwind styles defined directly in the content.

How is it included?

> since I don't know the styles ahead of time

Could it be possible to infer that, ahead of time?
Because in MDX we have ASTs! Markdown, HTML, and JavaScript ones.
So we are probably able to figure this out!

> I'm sure I got a bit ahead of myself since my use case is driven by content that starts as MDX [...]

Yeah of course, no problem, this plugin is a fine solution for you.
But: I am here thinking about all the folks that will see this plugin (and on the MDX site?), and what they’d expect `remark-mdx-postcss` (`remark-mdx-{css,style}`?) to do.

> Perhaps something like the following:

Yup! [`rehype-document`][https://github.com/rehypejs/rehype-document/blob/2b904e9aadaf1f5afbf9305d694cb82e4eb26713/index.js#L88-L93] does that.

> Do you think there's any issue with me trying to generalize here or do you recommend that these plugins try to handle specific use cases?

Depends! We could generalize in a different direction (as in, is this about tailwind, postcss, or css?)
Which way to go depends on a) what do you need, b) what do others need? c) what do you (or me/we, here in MDX/remark?) want to maintain?

wooorm:
Hey! Thanks for the kind words. I looked some more into tailwind, and it’s different now in v3 that I remember from v2.

> there's no guarantee that these arbitrary tailwind classes (i.e. `font-bold`) will be available for me to use as I create content.

This can only be guaranteed by shipping all of tailwind, right? But you are not shipping all of tailwind.

Q: what is this size of all-of-tailwind (minzipped)? Would it perhaps be easiest to just include tailwind on your website? Especially because stuff like `@tailwind base` and more will be duplicated now. Meaning there’s 6kb of CSS embedded for every MDX document! I find docs for v2 on [how to purge stuff](https://v2.tailwindcss.com/docs/optimizing-for-production) because it was giant, but for v3 [they removed those docs](https://tailwindcss.com/docs/optimizing-for-production) as I guess it’s an okay size?

I was guessing from this post that passing “purged” / “tree-shaken” code from tailwind was kinda your goal. But maybe that’s not needed?

SSHari:
So with the update of tailwind to v3 and their new JIT compiler, instead of purging styles, you define file paths for tailwind to scan and only the styles you use are generated: https://tailwindcss.com/docs/configuration#content.

I haven't done a test to see how large including all the styles would be, but that may be an option. With that said, your point about each MDX document containing 6kb of styles is fair, but we can get away from that by making some assumptions. Because the MDX document ultimately runs on my website which is also using tailwind, I can safely guarantee that the base styles will always be there, so I can skip those when generating styles for a given MDX document.

```css
@tailwind components;
@tailwind utilities;
```

vs

```css
@tailwind base;
@tailwind components;
@tailwind utilities;
```

There's still a chance for duplication because if I use something like `font-bold` in my MDX and that's already used in the main application, then the MDX is unnecessarily bundling that style. That said, the amount of CSS I'm putting into the MDX is really minor, so I'd really just end up with a style tag that looks something like:

```html
<style>.font-bold {font-weight: 700}.overline {text-decoration-line: overline}</style>
```

By leveraging tailwind's JIT compiler, I think I can end up with a situation where I'm bundling the absolute minimum with my main application and the MDX that's defined.

wooorm:
Given that you know what styles you need, at this point, might it not be easier to import stuff from tailwind from inside MDX files? Then it scales better: if one post needs a bunch of stuff, the other files aren’t affected?
E.g., what about such MDX:

```
export {styles} from 'tailwind/utilities'

# Hello, world!
```

Or:

```
<style>
  @tailwind base;
  @tailwind components;
  @tailwind utilities;
</style>

# Hello, world!
```

---

I see two other approaches: a) is a smart tailwind heavy solution, but then we need to ask tailwind folks (or for you to know the answer), because I can’t find docs. I think we’d want:
1. walk the MDX AST to find classes that are used in a document
2. optionally, if possible, walk the site tailwind css to exclude those styles
3. ask tailwind to generate CSS for `list 1. - list 2.`
4. inject the result in MDX files

Approach b) is changing this plugin to be `rehype-postcss` or a different name (this name [already exists](https://github.com/viktor-yakubiv/rehype-postcss)). So the focus is more: inject CSS and run postcss on it, to do whatever!

SSHari:
Ok! I think the two options you're putting forward make sense. Option A would probably leave me with the bare minimum I need for my use case to be successful. But to your point, we'd likely need to work with the tailwind team to get a better idea around how to accomplish this.

As for option B, I think that was my original intent. I took a look at `rehype-postcss` and it looks like I can actually leverage it for my use case as well. So long as I have the following `style` tag:

```html
<style>
  @tailwind components;
  @tailwind utilities;
</style>
```

I can just take my MDX content and pass that to the `tailwind` plugin the same way I was initially doing with the plugin I wrote. I think the use case we're talking about is actually solved with the plugin you've linked!

The one caveat is that the plugin only processes style tags that have the [element type](https://github.com/viktor-yakubiv/rehype-postcss/blob/main/index.js#L43). The problem is that if I drop a `style` tag in my MDX, it will have the type `mdxJsxFlowElement` and therefore won't be processed. I'm sure that there's no way to force the `element` type in this scenario because there's probably no way to make a distinction between JSX vs an HTML element in this case right?

The good news is that I can just add a rehype plugin locally that just adds the style tag above before passing it to the `rehype-postcss` plugin.

wooorm:
> I'm sure that there's no way to force the element type in this scenario because there's probably no way to make a distinction between JSX vs an HTML element in this case right?

It might be possible to have a JSX element that behaves like an HTML element. However, it’s probably all much easier with a rehype plugin, that instead of adding a JSX element, adds an HTML `<style>` element.

This could also become an option for `rehype-postcss`, to accept `options.style` as a `string` (or array of strings?) just like [`rehype-document` does](https://github.com/rehypejs/rehype-document#example-css)?

SSHari:
I think that's a good idea to be able to pass raw css as input to the `rehype-postcss` plugin. I can bring it up as a point of discussion in that repo. Regardless of whether a change like that makes it in, I think having a local rehype plugin that just injects a style tag before the `rehype-postcss` plugin will be good enough for the time being.

I'll go ahead and close this PR though since any potential changes could happen in the `rehype-postcss` repo.

Issue #1039, opened 2022-09-07
remark-parse/lib/index.d.ts imports a module that cannot be referenced


yamachu:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

remark-parse 10.0.1

### Steps to reproduce

https://github.com/yamachu/remark-types-broken-repro

1. Create a typescript project with strict type checking ( like this https://github.com/yamachu/remark-types-broken-repro/blob/699e5a1ef6963c6f0c6d06d503b962543ffb8945/tsconfig.json )
2. add dependencies unified 10.1.2 and remark-parse 10.0.1 
3. write code as below (code: REPRO)
4. execute `npx tsc --noEmit` for running type check

REPRO

```ts
import { unified } from "unified";
import markdown from "remark-parse";

const processor = unified().use(markdown);

const _ = processor.parse("**This is Markdown**");
```

### Expected behavior

Pass type check without error

### Actual behavior

Fail type check with error

### Runtime

Node v16

### Package manager

yarn 1

### OS

macOS

### Build and bundle tools

Other (please specify in steps to reproduce)

JounQin:
Can you paste some error logs?

yamachu:
thanks @JounQin 

here is error log
```
# ~/Projects/github.com/yamachu/remark-types-broken-repro $ [master]
 yarn run typecheck
yarn run v1.22.19
$ tsc --noEmit
node_modules/remark-parse/lib/index.d.ts:3:26 - error TS2307: Cannot find module 'mdast-util-from-markdown/lib' or its corresponding type declarations.

3   options: void | import('mdast-util-from-markdown/lib').Options | undefined
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/remark-parse/lib/index.d.ts:3

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

JounQin:
[...]

PR #1040, opened 2022-09-07
Rewrite import path to allow mdast-util-from-markdown type to resolve

yamachu:
### Initial checklist

### Description of changes

ref: https://github.com/remarkjs/remark/issues/1039

I edited the output type definition file because output of the TypeScript compiler cannot be controlled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "xo": "^0.50.0"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


JounQin:
I don't think it's a good solution, the issue is caused by TypeScript itself, so it should be fixed in TypeScript's side.

yamachu:
it means that the user should enable `skipLibCheck` ?

wooorm:
I agree with @JounQin here!

PR #1076, opened 2022-11-25
Add `remark-code-title` to list of plugins

kevinzunigacuellar:
Signed-off-by: Kevin Zuniga Cuellar <46791833+kevinzunigacuellar@users.noreply.github.com>

### Initial checklist

### Description of changes

Add [`remark-code-title`](https://github.com/kevinzunigacuellar/remark-code-title) plugin to list of plugins

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -298,6 +298,8 @@ The list of plugins:
     — new syntax for wiki links (rehype compatible)
 *   🟢 [`remark-yaml-config`](https://github.com/remarkjs/remark-yaml-config)
     — configure remark w/ YAML
+*   🟢 [`remark-code-title`](https://github.com/kevinzunigacuellar/remark-code-title)
+    — add titles to code blocks
 
 ## List of presets
 


remcohaszing (approved):

wooorm:
Hey! Congrats on making your plugin :)

I think it can be better though: you should *not* inject strings of HTML into the AST.
There is an AST for HTML already. hast. You can create a rehype plugin to work on the HTML AST. I’d recommend that.
If you really need to work on markdown, you should use the embedded HTML AST: https://github.com/syntax-tree/mdast-util-to-hast#fields-on-nodes.

The return value of `visit` is a) not needed, b) incorrect. If you specify it, it should I believe be `index + 1` instead.

kevinzunigacuellar:
Hello @wooorm, thanks a lot for taking the time to give some feedback. I agree, injecting strings feels a bit hacky. I will update this using the mdast-hast util. 

I will also explore the rehype plugin route but since the package is already published I will start that in a fresh repo. Again thanks a lot for the constructive feedback 🙌

kevinzunigacuellar:
hello @wooorm sorry to bother you again. I applied the requested changes except for the ` return index + 1`. When I changed this I ran into an infinite loop. I am not exactly sure what the reason behind this is, but I just wanted to let your know.

kevinzunigacuellar:
> Nice! Thanks for sharing.
> 
> An aside, for TypeScript + ESM, you may need `"type": "module"` in _package.json_, see https://2ality.com/2021/06/typescript-esm-nodejs.html for a guide on publishing ESM packages with TS.

Thanks, I completely missed that

kevinzunigacuellar:
> Nice! Thanks for sharing.
> 
> An aside, for TypeScript + ESM, you may need `"type": "module"` in _package.json_, see https://2ality.com/2021/06/typescript-esm-nodejs.html for a guide on publishing ESM packages with TS.

fixed in version `0.2.2` 

Thanks again for the feedback. I am very new to publishing npm packages.

wooorm (approved):

wooorm:
Thanks! :)

PR #1129, opened 2023-03-11
Add `remark-ins` to list of plugins

talatkuyuk:
### Initial checklist

### Description of changes

Added a new plugin `remark-ins` to list of plugins

https://github.com/ipikuka/remark-ins
https://www.npmjs.com/package/remark-ins

I've taken the experience which I got during development of my previous plugins.

Using the `remark-ins`, you can easily create \<ins\> element in markdown.

```markdown
++inserted text++
```
is going to produce:
```html
<p><ins>inserted text</ins></p>
```

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -163,6 +163,8 @@ The list of plugins:
     — prefix relative image paths with an absolute URL
 *   🟢 [`remark-inline-links`](https://github.com/remarkjs/remark-inline-links)
     — change references and definitions to links and images
+*   🟢 [`remark-ins`](https://github.com/ipikuka/remark-ins)
+    — add ins element for inserted texts opposite to deleted texts
 *   🟢 [`remark-join-cjk-lines`](https://github.com/purefun/remark-join-cjk-lines)
     — remove extra space between CJK Characters.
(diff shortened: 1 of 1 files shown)


remcohaszing (approved):
Thanks! This looks great!

Is this syntax based on something? Have you considered using [GLFM (GitLab flavoured markdown)](https://docs.gitlab.com/ee/user/markdown.html#inline-diff)?

ChristianMurphy (approved):

talatkuyuk:
> Thanks! This looks great!
> 
> Is this syntax based on something? Have you considered using [GLFM (GitLab flavoured markdown)](https://docs.gitlab.com/ee/user/markdown.html#inline-diff)?

Hi @remcohaszing,

Actually, I am trying to implement the syntax features from the https://github.com/markdown-it/markdown-it for the `remark` ecosystem. Their demo page is https://markdown-it.github.io

The syntax for the `inline diff` in `GLFM` is a bit different from inserted text with \<ins\> element. Is there a need to implement the `inline diff` syntax for the `remark`?

remcohaszing:
I think it’s best to implement existing syntaxes instead of inventing your own, as the latter would fragment the markdown ecosystem. I wasn’t aware of this syntax from `markdown-it`, but it appears you are implementing an existing syntax, which is great!

wooorm:
Thank you! :)

PR #1143, opened 2023-04-05
Add `remark-obsidian-callout` to list of plugins

escwxyz:
### Initial checklist

### Description of changes

This PR adds the plugin [`remark-obsidian-callout`](https://github.com/escwxyz/remark-obsidian-callout) to the list of plugins, also seen on [npm](https://www.npmjs.com/package/remark-obsidian-callout).

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -197,7 +197,7 @@ The list of plugins:
 *   🟢 [`@agentofuser/remark-oembed`](https://github.com/agentofuser/remark-oembed)
     — transform URLs for youtube, twitter, etc. embeds
 *   🟢 [`remark-obsidian-callout`](https://github.com/escwxyz/remark-obsidian-callout)
-    — parse Obsidian's callout syntax for further customizations
+    — parse Obsidian’s callout syntax for further customizations
 *   🟢 [`remark-oembed`](https://github.com/sergioramos/remark-oembed)
     — transform URLs surrounded by newlines into *asynchronously* loading
     embeds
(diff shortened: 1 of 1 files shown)


wooorm:
Hi there!

This should be a rehype plugin. rehype is for HTML. remark is for markdown. You work with HTML. So you should work on the HTML AST. Some more info on this can be found here: https://github.com/remarkjs/remark-rehype#what-is-this

remark/rehype/rest of unified are also about ASTs. So, you have an object that represents a paragraph, with children, for example.
You inject strings of HTML. Strings of HTML for the divs, strings of HTML for the icons. Injecting strings defeats the purpose of having ASTs.
There are many examples of rehype plugins that you can use as inspiration: https://github.com/rehypejs/rehype/blob/main/doc/plugins.md#list-of-plugins.
One that looks somewhat like what you want to do is: https://github.com/rehypejs/rehype-picture/blob/main/index.js

You also use TypeScript, but don’t use the types for the AST.
You can use specific types from `@types/mdast` / `@types/hast`, they subclass the abstract interfaces from `@types/unist` into interfaces that are actually used.
If you then **type** your plugin, instead of accepting `any`, you will find a much better user experience, because everything will be typed.
An example of how to type plugins is: https://github.com/remcohaszing/rehype-mdx-title/blob/6832b8c01097da83288319ddaa5d1e460d4e125a/index.ts#L23.

escwxyz:
> Other plugins can’t inspect them. They can’t be made safe. They don‘t work with MDX for example, or many other tools.

You're right, thanks for the detailed explanation and guidance! This was initially for the personal use with Astro, but as you have said, I would consider rewriting for rehype. Thanks!

wooorm:
Cool! :)

Let me know if you have any further questions, good luck!

PR #1164, opened 2023-05-17
add plugin `remark-shaku-code-annotate`

JSerZANP:
### Initial checklist

### Description of changes

Added `remark-shaku-code-annotate` to the plugin list. 

shaku-code-annotate-screenshot


wooorm:
Heya!

Your plugin does things with HTML.
remark is for markdown, rehype is for HTML. For more info, see each projects readme, or in short: https://github.com/remarkjs/remark-rehype#what-is-this.

You should create a rehype plugin instead of a remark plugin.
Now you are generating strings of HTML that we need to parse again, while the point of these projects is to use ASTs! Which is much quicker!

JSerZANP:
@wooorm 

I see,  thanks for the heads-up.
fyi, I'm following https://github.com/shikijs/twoslash/tree/main/packages/remark-shiki-twoslash

PR #1195, opened 2023-08-05
Use markdown-extensions

remcohaszing:
### Initial checklist

### Description of changes

I don’t know why it was problematic earlier, but `markdown-extensions` has just been changes to ESM itself and now has type definitions.

### DIFF
--- a/packages/remark-cli/cli.js
+++ b/packages/remark-cli/cli.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 import {createRequire} from 'node:module'
+import extensions from 'markdown-extensions'
 import {args} from 'unified-args'
 // eslint-disable-next-line import/order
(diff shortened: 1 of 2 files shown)


ChristianMurphy (approved):
Thanks @remcohaszing!

wooorm (approved):

wooorm:
My hunch is that when ESM started in Node, it couldn’t load JSON, even from ESM -> CJS -> JSON, but not sure anymore

wooorm:
Thanks, also for getting the types in there!

wooorm:
> I don’t know why it was problematic earlier,

I remembered: ESM can, as you know, import CJS. But, at least at the start, when using ESM here, it could not load JSON in that CJS. `markdown-extensions` was JSON, and failed.

PR #1197, opened 2023-08-06
Add `remark-d2` to list of plugins

mech-a:
### Initial checklist

### Description of changes

I've made a plugin, [remark-d2](https://github.com/mech-a/remark-d2), that makes it so you can turn `d2` code blocks into their compiled images. `d2` is a fast-to-develop and easy diagramming language, so I think it would be a good addition to the list. They also have a website where you can [compare d2 code to other diagramming tools](https://text-to-diagram.com/?b=mermaid) like Mermaid, PlantUML, and others.

I'm not affiliated with d2 in any way, in case that is of concern.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -88,6 +88,8 @@ The list of plugins:
 *   ⚠️ [`remark-custom-blocks`](https://github.com/zestedesavoir/zmarkdown/tree/HEAD/packages/remark-custom-blocks#readme)
     — new syntax for custom blocks (new node types, rehype compatible)
     (👉 **note**: [`remark-directive`][d] is similar and up to date)
+*   🟢 [`remark-d2`](https://github.com/mech-a/remark-d2)
+    — turn [d2](https://d2lang.com/) diagram code blocks into images
 *   🟢 [`remark-definition-list`](https://github.com/wataru-chocola/remark-definition-list)
     — support definition lists
 *   🟢 [`remark-defsplit`](https://github.com/remarkjs/remark-defsplit)


doc/plugins.md
```diff
@@ -88,6 +88,8 @@ The list of plugins:
 *   ⚠️ [`remark-custom-blocks`](https://github.com/zestedesavoir/zmarkdown/tree/HEAD/packages/remark-custom-blocks#readme)
     — new syntax for custom blocks (new node types, rehype compatible)
     (👉 **note**: [`remark-directive`][d] is similar and up to date)
+*   🟢 [`remark-d2`](https://github.com/mech-a/remark-d2)
+    — turn [d2](https://d2lang.com/) diagram code blocks into images
```
wooorm, line 92: Hi! Some tips on how to improve your plugin:

- You don’t need `'use strict'` in ESM, they are always strict
- Instead of using `node:path`, I recommend using URLs, they work across platforms (unix vs windows)
- Throw errors instead of printing to `console.error` when things are misconfigured
- You don’t need to `path.parse(file.path)`, `file` already has `direname` and other fields
- replace `process.cwd()` with `file.cwd`, files can theoretically exist with multiple, different, “`cwd`s”
- Use async versions of `fs` calls, the transformer can be async
- an `exists` call before a `mkdir` or `write` call is an anti-pattern, instead, try to `mkdir` or `write`, and swallow the error if it already exists
- do not change `node.type`, instead, replace one node with another, some more info here: https://unifiedjs.com/learn/recipe/remove-node/#removing-a-node
- It is probably better to do async work instead of `spawn`, so that when many files are processed, or many code blocks exist, work can be done at the same time
- What is the benefit of the HTML version? What can it do more than the markdown version? If there are no benefits, I recommend removing it, because raw strings of HTML don’t always work well with markdown, particularly when dealing with VDOM frameworks like React, such as MDX. There are also perhaps openings for XSS attacks, or at least bugs, when building a string of HTML like you do now, instead of using something like [hastscript](https://github.com/syntax-tree/hastscript) and [hast-util-to-html](https://github.com/syntax-tree/hast-util-to-html)!

mech-a: Thank you for this feedback. I'll integrate your feedback where applicable and reply back.

wooorm: how’s it going with the feedback?

wooorm: Ping! :)

PR #1248, opened 2023-10-24
Add `remark-hexo` to list of plugins

bennycode:
### Initial checklist

### Description of changes

I added a link to the [remark-hexo](https://github.com/bennycode/remark-hexo) plugin which I built.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -67,6 +67,8 @@ The list of plugins:
     — populate code blocks from files
 *   🟢 [`remark-code-screenshot`](https://github.com/Swizec/remark-code-screenshot)
     – turn code blocks into `carbon.now.sh` screenshots
+*   🟢 [`remark-code-title`](https://github.com/kevinzunigacuellar/remark-code-title)
+    — add titles to code blocks
 *   🟢 [`remark-codesandbox`](https://github.com/kevin940726/remark-codesandbox)
     – create CodeSandbox from code blocks
 *   🟢 [`remark-collapse`](https://github.com/Rokt33r/remark-collapse)
@@ -146,6 +148,8 @@ The list of plugins:
     — serialize w/ more blank lines between headings
 *   🟢 [`@vcarl/remark-headings`](https://github.com/vcarl/remark-headings)
     — extract a list of headings as data
+*   🟢 [`remark-hexo`](https://github.com/bennycode/remark-hexo)
+    — renders [Hexo tags](https://hexo.io/docs/tag-plugins)
 *   🟢 [`remark-highlight.js`](https://github.com/remarkjs/remark-highlight.js)
     — highlight code blocks w/ [highlight.js](https://github.com/isagalaev/highlight.js)
     (rehype compatible)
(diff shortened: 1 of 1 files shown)


doc/plugins.md
```diff
@@ -316,6 +316,8 @@ The list of plugins:
     — configure remark w/ YAML
 *   🟢 [`remark-code-title`](https://github.com/kevinzunigacuellar/remark-code-title)
     — add titles to code blocks
+*   🟢 [`remark-hexo`](https://github.com/bennycode/remark-hexo)
+    — renders [Hexo tags](https://hexo.io/docs/tag-plugins)
```
wooorm, line 320: Can you sort it in the list? Thanks

(I see the one above is also wrongly places, meh, anyway, feel free to ignore that!)

bennycode: No problem! I sorted mine as well as the one from above (remark-code-title) :)

ChristianMurphy:
Thanks @bennycode!
A few thoughts:
1. running this as a transformer, creates a caveat that it may not work with other plugins which look for the `{` symbol (like MDX). If that is needed, you may need to parse at the micromark level: see https://github.com/micromark/micromark#extending-markdown and https://github.com/micromark/micromark#creating-a-micromark-extension
2. If it continues as a transformer, consider leveraging https://github.com/syntax-tree/mdast-util-find-and-replace
3. Is it intentional that the code does not have a license? Without a license it is not open source and others cannot use it. Consider a permissive license like MIT or Apache 2 for more adoption, or GPL or SSPL for stronger copyleft.
4. In your tsconfig, consider setting module resolution to `node16` to ensure compatibility with TypeScript with ESM

ChristianMurphy:
> What is so wrong about my technique? 

It's not necessarily "wrong".
I intuit that there may be some edge cases you are not considering:

1. what happens when `remark-hexo` is used with other parser plugins like `remark-mdx`? `remark-mdx` is a parser plugin, it runs at the very beginning, when it sees `{` it inteprets it as a JSX block, it does not know about `remark-hexo` and will fail the document because `{%` is invalid JSX. Alternatively, you can document "this may not work with `remark-mdx` or other plugins that use `{}`".
2. There could be text before or after the Hexo template, currently that text will be erased/replaced. `mdast-util-find-and-replace` will keep the other surrounding text. (you could also do this with `unist-util-visit` but it is more involved)

bennycode:
Ah, thanks for the detailed info. I am using the `remark-hexo` plugin myself in an Astro blog (which uses Remark & Rehype) and is it was working fine I was assuming there is nothing wrong with it. Need to double check what happens with MDX files as I was only using the `{` syntax in MD files.

ChristianMurphy:
Again, you don't _have_ to make them work together.
MDX and Hexo feel more like alternatives than something to use together.
But it could be something to document as an intended incompatibility to avoid people logging issues, since MDX is something commonly used with Astro.

remcohaszing (approved):

wooorm (approved):

wooorm:
Thanks!

PR #1281, opened 2024-01-23
Add `remark-steps` to list of plugins

lituidev:
### Initial checklist

### Description of changes

ADD

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -265,6 +265,8 @@ The list of plugins:
     — reorder reference-style link definitions
 *   🟢 [`remark-sources`](https://github.com/unlight/remark-sources)
     — insert source code
+*   🟢 [`remark-steps`](https://github.com/lit-ui/remark-steps)
+    — Rebuild steps style into DOM
 *   🟢 [`remark-strip-badges`](https://github.com/remarkjs/remark-strip-badges)
     — remove badges (such as `shields.io`)
 *   🟢 [`remark-strip-html`](https://github.com/craftzdog/remark-strip-html)


ChristianMurphy:
Welcome @lituidev! 👋 
Thanks for sharing!
A few thoughts:
1. This plugin appears to be a specific kind of directive, it would be good to note both in the description here and in the package itself that directives are required (source: https://github.com/lit-ui/remark-steps/blob/ad235a49c593a5b57e3308552844c89f1d3b2d6e/src/index.ts#L23)
2. There is a types only package for MDAST nodes https://www.npmjs.com/package/@types/mdast (avoid hard coding your own node type, source: https://github.com/lit-ui/remark-steps/blob/ad235a49c593a5b57e3308552844c89f1d3b2d6e/src/index.ts#L12-L19)
3. HAST and MDAST are not the same, avoid pushing HAST nodes as children of MDAST, instead use `data.hChildren` https://github.com/syntax-tree/mdast-util-to-hast#fields-on-nodes (source: https://github.com/lit-ui/remark-steps/blob/ad235a49c593a5b57e3308552844c89f1d3b2d6e/src/index.ts#L32-L42)
4. You probably want to target `node16` module resolution for ESM https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext (source: https://github.com/lit-ui/remark-steps/blob/ad235a49c593a5b57e3308552844c89f1d3b2d6e/tsconfig.json#L4-L6)
5. You probably want to publish somewhat readable source, let downstream adopters decide how they want to transpile, polyfill, or minify code (source: https://github.com/lit-ui/remark-steps/blob/ad235a49c593a5b57e3308552844c89f1d3b2d6e/rollup.config.mjs#L17)

wooorm:
Closing this as it seems abandoned. This also makes more sense as a rehype plugin I think!

PR #1293, opened 2024-03-03
Add `remark-github-admonitions-to-directives` to list of plugins

LuudJanssen:
### Initial checklist

### Description of changes

I created a plugin to turn GitHub's alert syntax to the directives syntax used by admonitions. I'm curious to hear what you think and if it fits the list 😊

Plugin: https://github.com/incentro-dc/remark-github-admonitions-to-directives

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -138,6 +138,8 @@ The list of plugins:
     — add a table of contributors based on Git history, options, and more
 *   🟢 [`remark-github`](https://github.com/remarkjs/remark-github)
     — autolink references to commits, issues, pull-requests, and users
+*   🟢 [`remark-github-admonitions-to-directives`](https://github.com/incentro-dc/remark-github-admonitions-to-directives)
+    — convert GitHub's blockquote-based admonitions syntax to directives syntax
 *   🟢 [`remark-github-beta-blockquote-admonitions`](https://github.com/myl7/remark-github-beta-blockquote-admonitions)
     — [GitHub beta blockquote-based admonitions](https://github.com/github/feedback/discussions/16925)
 *   ⚠️ [`remark-grid-tables`](https://github.com/zestedesavoir/zmarkdown/tree/HEAD/packages/remark-grid-tables#readme)


ChristianMurphy (approved):
Thanks for sharing @LuudJanssen!

Some thoughts:
1. consider annotating the plugin with `Plugin<[], Root>` to specific the node type it accepts and returns https://github.com/incentro-dc/remark-github-admonitions-to-directives/blob/6c2915ec62b252752c5e1ddff164d1160b5e5f9b/src/plugin.ts#L8
2. `type` exports need to be put ahead of any other exports in package.json docs: https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports source: https://github.com/incentro-dc/remark-github-admonitions-to-directives/blob/6c2915ec62b252752c5e1ddff164d1160b5e5f9b/package.json#L24-L30
3. Plug and play module systems (pnpm and yarn) need types to be specified as `dependencies` https://github.com/incentro-dc/remark-github-admonitions-to-directives/blob/6c2915ec62b252752c5e1ddff164d1160b5e5f9b/src/parse-github-alert-blockquote.ts#L1
4. Any system which doesn't recognize `exports` will break when using `main`, as there are no `.jss` files, only `.js` files https://github.com/incentro-dc/remark-github-admonitions-to-directives/blob/6c2915ec62b252752c5e1ddff164d1160b5e5f9b/package.json#L31

LuudJanssen:
Hi @ChristianMurphy, thank you so much for your awesome suggestions. I've applied all of them and released a v1.0.1 with the changes.

ChristianMurphy (approved):
Looks good!

Very minor, I'd recommend widening version ranges of dependencies to the full major range https://github.com/incentro-dc/remark-github-admonitions-to-directives/blob/f67299024f9b9b58f4250a00300433b6a1f13f21/package.json#L48-L50
E.G. `"@types/mdast": "^4.0.0",` this allows package managers to deduplicate more.
I've seen more issues from confusion on having multiple copies of the same library living on a computer than from being a patch or two behind.

LuudJanssen:
Hi @ChristianMurphy, great suggestion. Even though the package manager should resolve this for most cases, there's always that dependency that has a fixed patch version. I've loosened up the version ranges and released [v1.0.4](https://github.com/incentro-dc/remark-github-admonitions-to-directives/releases/tag/1.0.4).

wooorm (approved):

wooorm:
Cool stuff Luud!

PR #1371, opened 2024-09-25
feat: add star count for each plugin

nikitarevenco:
### Initial checklist

* [x] I read the support docs 
* [x] I searched issues and couldn’t find anything (or linked relevant results below) 
* [ ] If applicable, I’ve added docs and tests

### Description of changes

Stars help me quickly filter the plugins to decide which ones I want to try out, and this PR adds that for every plugin!

Thankfully we have AI otherwise this would take me like 2 hours :smile:

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -32,310 +32,147 @@ More plugins can be found on GitHub tagged with the
 
 The list of plugins:
 
-* [`remark-a11y-emoji`](https://github.com/florianeckerstorfer/remark-a11y-emoji)
-  — accessible emoji
-* ⚠️ [`remark-abbr`](https://github.com/zestedesavoir/zmarkdown/tree/HEAD/packages/remark-abbr#readme)
-  — new syntax for abbreviations (new node type, rehype compatible)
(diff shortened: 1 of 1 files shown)


wooorm:
Heya! I think there’s sometimes something to say for data such as this. Quantitative. And you can use GH to do that: you can [search for topics](https://github.com/topics/remark-plugin), or a [general search](https://github.com/search?q=remark&type=repositories&s=stars&o=desc).

But there’s also something for without. Qualitative. And that’s this list. A list we maintain. Where people can add things. And we look at whether it’s a good and recommended approach.

There are very good plugins with very few stars.
There are pretty bad plugins with many stars.

I’d prefer this list without the stars. /cc @remcohaszing @christianmurphy

remcohaszing:
I kind of like the idea, but I agree with @wooorm’s points.

Also the list looks visually less appealing to me with these badges.

nikitarevenco:
Ok, I can see where you guys are coming from!

ChristianMurphy:
I'm neutral on adding stars or other stats like download count.
If we did add these, the plugins section should be reworked as a table.

PR #1375, opened 2024-10-02
Add remark-prepend-url to list of plugins

alxjpzmn:
### Initial checklist

* [x] I read the support docs 
* [x] I searched issues and couldn’t find anything (or linked relevant results below) 
* [x] If applicable, I’ve added docs and tests

### Description of changes

Add [remark-prepend-url](https://github.com/alxjpzmn/remark-prepend-url) to the plugin list.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -233,6 +233,8 @@ The list of plugins:
 * ⚠️ [`remark-ping`](https://github.com/zestedesavoir/zmarkdown/tree/HEAD/packages/remark-ping#readme)
   — new syntax for mentions w/ configurable existence check (new node
   type, rehype compatible)
+* 🟢 [`remark-prepend-url`](https://github.com/alxjpzmn/remark-prepend-url)
+  —  prepend an absolute url to relative links
 * 🟢 [`remark-prettier`](https://github.com/remcohaszing/remark-prettier)
   — check and format markdown using Prettier
 * 🟢 [`remark-prism`](https://github.com/sergioramos/remark-prism)


ChristianMurphy:
Welcome @alxjpzmn 👋 
Thanks for sharing

The types look a bit off, you may want to leverage `@types/mdast`
https://github.com/alxjpzmn/remark-prepend-url/blob/ab819c73f4c4dd01d044c8da2679c6cc8531b272/index.ts#L9-L10

The plugin will be passed a `Root` node, and once that is in place, your visit function should be able to infer it will be given an `Link` node without additional annotations.
See https://unifiedjs.com/learn/recipe/tree-traversal-typescript/ for reference

> Useful if you're using a static site generator like Astro with .mdx files and you'd like to prepare them for a full-text RSS feed.

Is the intent that you want to modify the source markdown to have the absolute URL? (if so a remark plugin is good)
If you only want it to change the resulting HTML, you may want a rehype plugin.

alxjpzmn:
Thanks for the quick review @ChristianMurphy & the pointer re: the types :)

Changed to `@types/mdast` and published a new release [here](https://github.com/alxjpzmn/remark-prepend-url/compare/1.0.0...1.0.1).

> Is the intent that you want to modify the source markdown to have the absolute URL? (if so a remark plugin is good) 

For my personal use case (generating a XML file) rehype would work, but I think being able to modify the source markdown is more versatile. For example, if I author a markdown file with relative links and the desired output is not only HTML but also a e.g. PDF, I'd first want the source markdown to uniformly contain absolute links.

ChristianMurphy (approved):
LGTM!

remcohaszing (approved):

wooorm:
Thank you!

PR #1406, opened 2025-02-10
Update plugins.md adding remark-callouts and remark-wiki-link-plus

rufuspollock:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Update plugins.md adding remark-callouts and remark-wiki-link-plus

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -50,6 +50,8 @@ The list of plugins:
   — increase or decrease heading depth
 * 🟢 [`remark-breaks`](https://github.com/remarkjs/remark-breaks)
   – hard breaks w/o needing spaces (like on issues)
+* 🟢 [`remark-callouts`](https://github.com/flowershow/remark-callouts)
+  - Remark plugin to add support for blockquote-based callouts/admonitions similar to the approach of Obsidian and Microsoft Learn style.
 * 🟢 [`remark-capitalize`](https://github.com/zeit/remark-capitalize)
   – transform all titles w/ [`title.sh`](https://github.com/zeit/title)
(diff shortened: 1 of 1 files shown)


wooorm (commented):
Hi!

CI is failing :)

Looks like the latter is deprecated? Both seem not really maintained?

I also see a `.use(markdown, { gfm: true })`, I don’t believe that option exists, it used to be there years and years ago.

For the callouts, that has more to do with HTML than with markdown. That’s where *rehype* comes in, instead of *remark*. Solutions like https://github.com/rehypejs/rehype-github/tree/main/packages/alert and https://github.com/chrisweb/rehype-github-alerts are good examples of that

rufuspollock:
> Looks like the latter is deprecated? Both seem not really maintained?

We actively maintain them - we just need to migrate some of the material back from the monorepo we created. We'll do that and update here.

> I also see a .use(markdown, { gfm: true }), I don’t believe that option exists, it used to be there years and years ago.

We'll also remove that.

> For the callouts, that has more to do with HTML than with markdown. That’s where rehype comes in, instead of remark. Solutions like https://github.com/rehypejs/rehype-github/tree/main/packages/alert and https://github.com/chrisweb/rehype-github-alerts are good examples of that

Ok, that's good to know and we'll look at that.

🙏 thank-you for prompt and clear response (as an open source maintainers ourselves we appreciate the time and effort!)

wooorm:
OK, well, no rush. I’ll close this for now then; you can reopen when you merge things back from the monorepo into separate things. Though, this list *can* have packages in monorepos, so that migration is not needed for this list.

This list is for plugins that we recommend to users. So that means running tests in CI and lots of tests and extensive docs and such!

Feel free to ask a maintainer here if you want someone to look through your code.
In many cases, we also recommend *rehype* plugins instead of *remark* plugins.

PR #1410, opened 2025-02-19
Add `remark-merge-data` to list of plugins

s-h-a-d-o-w:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Adds [`remark-merge-data`](https://github.com/s-h-a-d-o-w/remark-merge-data) to list of plugins.

One use case for this is to share configuration across chart definitions when creating charts using code, like in my article here (which uses [`remark-kroki`](https://github.com/show-docs/remark-kroki)): https://aop.software/blog/2025-01-31_evaluating-llrt/ ([code that makes use of `remark-merge-data`](https://github.com/s-h-a-d-o-w/blog/blob/master/astro.config.ts#L31-L140))

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -227,6 +227,8 @@ The list of plugins:
   — support MDX (JSX, expressions, ESM)
 * 🟢 [`remark-mentions`](https://github.com/FinnRG/remark-mentions)
   — replace @ mentions with links
+* 🟢 [`remark-merge-data`](https://github.com/s-h-a-d-o-w/remark-merge-data)
+  — merge globally defined data with data declared across code blocks
 * 🟢 [`remark-mermaidjs`](https://github.com/remcohaszing/remark-mermaidjs)
   — transform mermaid code blocks into inline SVGs
 * 🟢 [`remark-message-control`](https://github.com/remarkjs/remark-message-control)


ChristianMurphy:
Thanks for sharing @s-h-a-d-o-w!
Could you expand the readme a bit?
Perhaps including the markdown file it's intended to work with? Currently I don't really understand from the current docs what this is or why an adopter would use it.

s-h-a-d-o-w:
Thanks for the feedback! I agree that the README wasn't sufficiently clear. I wrote that while being very focused on my use case, so I didn't notice that it's lacking context.

What do you think? https://github.com/s-h-a-d-o-w/remark-merge-data/blob/36cb97274229e69073037eb1744f2ab42c7c33d7/README.md

ChristianMurphy:
The additional docs help!
It looks like they're on a fork? Not in the source of the plugin being added here.

s-h-a-d-o-w:
Nope, that was a PR. I waited for your approval before merging and releasing it. Which I have done just now.

ChristianMurphy (approved):
Looks good!
A few thoughts
* This check _may_ not be needed https://github.com/s-h-a-d-o-w/remark-merge-data/blob/e996968112056045034a5b9a1bb14157aa80fb20/src/index.ts#L43-L48, `isYaml` could be used to create a discriminated union https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions, which would allow this to be split https://github.com/s-h-a-d-o-w/remark-merge-data/blob/e996968112056045034a5b9a1bb14157aa80fb20/src/index.ts#L7-L30 and TypeScript to give an error if the incorrect data type is passed in
* Would data truly be unknown? or could it be constrained to be JSON? https://github.com/sindresorhus/type-fest?tab=readme-ov-file#json
* https://github.com/s-h-a-d-o-w/remark-merge-data/blob/e996968112056045034a5b9a1bb14157aa80fb20/src/index.ts#L63-L66 This will only support hard-coded meta attributes, if you also want to support dynamic, you may want to look to see if https://github.com/remcohaszing/rehype-mdx-code-props#readme could be supported as well

s-h-a-d-o-w:
> from your example, it seems its focus is generating HTML?

How so?

ChristianMurphy:
The main example you link to as motivating this is about generating rehype/HTML output https://aop.software/blog/2025-02-17_graphs-in-blogs/
And for Kroki or any other tool to work they generally need to be run through a renderer, presumably as part of the process you are doing.

s-h-a-d-o-w:
But the HTML output is generated by other plugins. This plugin only serves to share global data that is eventually used by those plugins.
(This just made me improve the example in the README though, to illustrate the data sharing aspect better. Maybe this makes it a bit clearer.)

PR #1411, opened 2025-02-19
Add `remark-kroki` to list of plugins

s-h-a-d-o-w:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Adds [`remark-kroki`](https://github.com/show-docs/remark-kroki) to list of plugins.

Transforms [kroki](https://github.com/yuzutech/kroki) code blocks into diagrams.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -211,6 +211,8 @@ The list of plugins:
 * ⚠️ [`remark-kbd-plus`](https://github.com/twardoch/remark-kbd-plus)
   — new syntax for keyboard keys w/ plusses (new node type, rehype
   compatible)
+* 🟢 [`remark-kroki`](https://github.com/show-docs/remark-kroki)
+  — transform kroki code blocks into diagrams
 * 🟢 [`remark-license`](https://github.com/remarkjs/remark-license)
   — add a license section
 * 🟢 [`remark-link-rewrite`](https://github.com/rjanjic/remark-link-rewrite)


ChristianMurphy:
Thanks for sharing @s-h-a-d-o-w!
Some suggestions:
1. it looks like the tests are not automatically run or checked by CI/CD, they should be run
2. There are no typescript types provided, those can be added to JS using JSDoc comments, without needed to convert to .ts files https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html and should be published to npm
3. The release does not appear to be labelled as semver stable https://github.com/show-docs/remark-kroki/blob/5d2f4984c9bcbee796b319978e42c6e31dd95488/package.json#L3 is this ready for outside consumption? If so could it release as 1.0.0?
4. The AST appear to use MDX specific constructs https://github.com/show-docs/remark-kroki/blob/5d2f4984c9bcbee796b319978e42c6e31dd95488/lib/transform.mjs#L28 but does not mention MDX in the readme
5. The module is published with `type: module` https://github.com/show-docs/remark-kroki/blob/5d2f4984c9bcbee796b319978e42c6e31dd95488/package.json#L43 you can use plain `.js` extensions, `.mjs` is not needed.

s-h-a-d-o-w:
Huh, um... I'm actually not maintaining that plugin and so wasn't aware of any but the most obvious of these points.

The MDX aspect in particular now has me wonder whether the advice I've given [in my article](https://aop.software/blog/2025-02-17_graphs-in-blogs/) could actually be problematic for many that might try it. I don't know how common blogs that don't use MDX are...

But so I unfortunately also can't easily change any of these things and don't have the time to start these discussions with the actual owner.

wooorm:
This may also probably be a rehype plugin? I think it doesn’t need “html” vs “mdx3” targets then.

The suggestion on using [`rehype-raw`](https://github.com/show-docs/remark-kroki#troubleshooting), when using the MDX mode, could also be solved, especially as a rehype plugin, by using [`hast-util-from-html-isomorphic`](https://github.com/syntax-tree/hast-util-from-html-isomorphic).

remcohaszing:
I agree with @wooorm. This is an HTML transform and would be best solved as a rehype plugin.

I don’t think we should recommend this remark plugin.

s-h-a-d-o-w:
Thanks for all the feedback. I'm not familiar with rehype tools yet but hopefully, I'll get around to migrating that project to rehype at some point. Because I think that the flexibility that kroki provides in terms of creating charts and whatnot is fantastic.

Issue #1288, opened 2024-02-13
Incorrect Parsing of Markdown with Special Characters Adjacent to Strong Tags in Non-English Contexts


YvesRijckaert:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

"remark-gfm": "^4.0.0", "remark-parse": "^11.0.0", "unified": "^11.0.4"

### Steps to reproduce

- Use `remark-parse`, `remark-gfm`, and `unified` packages to parse markdown content.
```
const processor = unified().use(remarkParse).use(remarkGfm);
const tree = processor.parse(md);
console.log(tree.children);
```
- Parse these two strings to see the different behaviour:
```
const markdownWithSpecialCharacter = '**test:**如李施德林';
const markdownWithoutSpecialCharacter = '**test**如李施德林';
```
- Observe that the raw markdown (`**`) is displayed instead of actual bold/strong text in the output, when there is a special character (here: `:`).

markdownWithSpecialCharacter:
```
[
  {
    "type": "paragraph",
    "children": [
      {
        "type": "text",
        "value": "**test:**如李施德林",
        "position": {
          "start": { "line": 1, "column": 1, "offset": 0 },
          "end": { "line": 1, "column": 15, "offset": 14 }
        }
      }
    ],
    "position": {
      "start": { "line": 1, "column": 1, "offset": 0 },
      "end": { "line": 1, "column": 15, "offset": 14 }
    }
  }
]
```

markdownWithoutSpecialCharacter:
```
[
  {
    "type": "paragraph",
    "children": [
      {
        "type": "strong",
        "children": [
          {
            "type": "text",
            "value": "test",
            "position": {
              "start": { "line": 1, "column": 3, "offset": 2 },
              "end": { "line": 1, "column": 7, "offset": 6 }
            }
          }
        ],
        "position": {
          "start": { "line": 1, "column": 1, "offset": 0 },
          "end": { "line": 1, "column": 9, "offset": 8 }
        }
      },
      {
        "type": "text",
        "value": "如李施德林",
[...]

Issue #1107, opened 2023-01-18
In Japanese, Emphasis and Strong have a bug


Neos21:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

remark-parse@7.0.1, remark-rehype@8.0.0

### Steps to reproduce

I use Node.js v18.12.1 and npm v8.19.2. I wrote vanilla Node.js script.

- `package.json` : `devDependencies`

```json
{
  "rehype-stringify": "9.0.3",
  "remark-parse": "10.0.1",
  "remark-rehype": "10.1.0",
  "unified": "10.1.2"
}
```

- Code:

```javascript
import { unified } from 'unified';
import remarkParse from 'remark-parse';
import remarkRehype from 'remark-rehype';
import rehypeStringify from 'rehype-stringify';

const processor = unified()
  .use(remarkParse)
  .use(remarkRehype)
  .use(rehypeStringify);
const result = processor.processSync(`
1. 日本語*強調*、日本語
2. 日本語*強調*。日本語
3. 日本語*強調、*日本語
4. 日本語*強調。*日本語
5. 日本語**強調**、日本語
6. 日本語**強調**。日本語
7. 日本語**強調、**日本語
8. 日本語**強調。**日本語
`);
console.log(result.value);
```

- 8 examples.
- The first 4 examples are wanted to convert to `<em>`
- The last 4 examples are wanted to convert to `<strong>`
- But, No. 3, 4, and 7, 8 were not converted.
    - I think a regexp has a bug with the character `、` and `。` maybe...?

### Expected behavior

```html
<ol>
<li>日本語<em>強調</em>、日本語</li>
<li>日本語<em>強調</em>。日本語</li>
<li>日本語<em>強調、</em>日本語</li>
<li>日本語<em>強調。</em>日本語</li>
<li>日本語<strong>強調</strong>、日本語</li>
<li>日本語<strong>強調</strong>。日本語</li>
<li>日本語<strong>強調、</strong>日本語</li>
<li>日本語<strong>強調。</strong>日本語</li>
[...]

PR #1422, opened 2025-03-15
Introduce `remark-cjk-friendly(-gfm-strikethrough)` in plugins.md

tats-u:
### Initial checklist

* [x] I read the support docs 
* ~~I included tests (or that’s not needed)~~ (N/A)

### Description of changes

Adds the following plugins:

- [`remark-cjk-friendly`](https://www.npmjs.com/package/remark-cjk-friendly)
- [`remark-cjk-friendly-gfm-strikethrough`](https://www.npmjs.com/package/remark-cjk-friendly-gfm-strikethrough)

[CommonMark's emphasis specification is unfriendly to CJK languages](https://github.com/commonmark/commonmark-spec/issues/650). (c.f. #1107 and #1288). These plugins mitigate this problem and are served as a public test for a new Commonmark specification candidate.

repo: https://github.com/tats-u/markdown-cjk-friendly/


ChristianMurphy:
Thanks for sharing @tats-u! 🙇 
A few notes:
1. it looks like the types are intended to be built https://github.com/tats-u/markdown-cjk-friendly/blob/b9c2ac657a0441ca47938a565491abe69b777e81/packages/remark-cjk-friendly-gfm-strikethrough/package.json#L5-L7 but didn't actually make it to npm https://www.npmjs.com/package/remark-cjk-friendly-gfm-strikethrough (in a way npm understands)
2. The module type should probably be `node16` or `nodenext` https://github.com/tats-u/markdown-cjk-friendly/blob/b9c2ac657a0441ca47938a565491abe69b777e81/packages/remark-cjk-friendly-gfm-strikethrough/tsconfig.json#L3-L4 source https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-node18-nodenext
3. I'd recommend against bundling as part of build https://github.com/tats-u/markdown-cjk-friendly/blob/b9c2ac657a0441ca47938a565491abe69b777e81/packages/remark-cjk-friendly-gfm-strikethrough/package.json#L32 this introduces indirection that makes the code more complex, less performant, and harder to debug. Ship code that has the types removed, but otherwise is as close as possible to the source.

tats-u:
> 2.

I adopted it only for libraries because the language server recommends `"moduleResolution": "nodenext"` for libraries and `"bundlers"` only for applications.

> 1.

There are .d.ts files in NPM:

https://www.npmjs.com/package/remark-cjk-friendly?activeTab=code / https://www.npmjs.com/package/remark-cjk-friendly-gfm-strikethrough?activeTab=code

> 3.

There is only single .js file in both remark plugin packages. https://www.npmjs.com/package/remark-cjk-friendly?activeTab=code / https://www.npmjs.com/package/remark-cjk-friendly-gfm-strikethrough?activeTab=code

> makes the code more complex, less performant, and harder to debug

The structure of the code is mostly kept except for the long prefix `__WEBPACK_EXTERNAL_MODULE_` like The Nile or The Great Wall. If you hate it, I recommend you to file an issue in Rslib.

tats-u:
My remark plugins only register corresponding micromark plugins as remark-gfm does.

ChristianMurphy:
> I adopted it only for libraries

Your code is a library.
It is published to npm, not as a web application.
Node16 or nodenext is correct.
To quote the typescript docs

> A common misconception is that node16—nodenext only emit ES modules. In reality, these modes describe versions of Node.js that support ES modules, not just projects that use ES modules. Both ESM and CommonJS emit are supported, based on the [detected module format](https://www.typescriptlang.org/docs/handbook/modules/reference.html#module-format-detection) of each file. Because they are the only module options that reflect the complexities of Node.js’s dual module system, **they are the only correct module options for all apps and libraries that are intended to run in Node.js v12 or later, whether they use ES modules or not.**

https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-node18-nodenext

> There are .d.ts files in NPM:

NPM does not recognize them, it expects a type property https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package
Which means your module is flagged as being untyped regardless of having .d.ts files

> If you hate it, I recommend you to file an issue in Rslib.

They already support it with `bundle: false`. https://lib.rsbuild.dev/config/lib/bundle#bundle-false
This isn't an rslib issue it is configuration issue in your code.

ChristianMurphy:
> My remark plugins only register corresponding micromark extensions as remark-gfm does.

I'm commenting on the remark plugins, because you are asking the remark team to recommend them.
That said my comments also apply to the micromark and markdown it code in the same repo.

tats-u:
> It is published to npm, not as a web application.

My repo contains demo applications for Docusaurus, Rspack, and Astro. Their tsconfig has not changed.

> [!IMPORTANT]
> tsconfig is not bundles into my packages. It is used only when built by Rslib.

https://github.com/tats-u/markdown-cjk-friendly/commit/968978b3385ad71cc80082c693185b60b8c5083c

> They already support it with bundle: false. 

I tried it but it didn't erase the super long variable prefix. It doesn't have effect if the number of .ts file is just one.

ChristianMurphy:
> My repo contains demo applications for Docusaurus, Rspack, and Astro. Their tsconfig has not been changed.

1. Everything used with node12+ should use the correct module specifier according to the typescript team themselves. Feel free to log bugs with other tools if they don't follow conventional.
2. The code is published to npm should follow library conventional, not app ones.

> I tried it but it didn't erase the super long variable prefix. It doesn't have effect if the number of .ts file is just one.

You may not need rslib at all.
Right now it's only acting as an unnecessary wrapper around the typescript compiler.
You could just call the typescript compiler directly.

ChristianMurphy:
> tsconfig is not bundles into my packages. It is used only when built by Rslib.

Right, when it is used by the typescript compiler it is producing a potentially wrong output.

tats-u:
> it expects a type property https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

Prettier doesn't have the top-level `"types"` but recognized as typed because `"types"` exists in every entry in `"export"`. My package.json adopts the same strategy because my packages don't need to support Node versions that don't recognize `"export"`. However, since only micromark-extension-cjk-friendly-gfm-friendly, which has the top-level `"types"`, has the TS badge, so I chose to add the top-level `"types"` to every maintained package.

https://www.npmjs.com/package/prettier?activeTab=code

> markdown it code

markdown-it plugins are out of the unified ecosystem. I won't turn off bundling it. (in the first place `bundle: false` doesn't have effect in ESM output there)

> Right now it's only acting as an unnecessary wrapper around the typescript compiler.

If I didn't use `v` flag of the regex, I would have to transpile the cutting edge syntax using Rslib or similar software.

> You may not need rslib at all.

Many libraries use [tsup](https://tsup.egoist.dev/) and I recognize Rslib plays the same role. I've started trying it, but the output looks better.

ChristianMurphy:
> Prettier doesn't have the top-level "types" but recognized as typed because "types" exists in every entry in "export"

That works for actual code.
But for npm to show the typescript badge, it's not enough.
I think it shows for them because they have index.d.ts at the root level.

> I chose to add the top-level "types" to every maintained package.

Sounds good.

> the cutting edge syntax using Rslib or similar software.

Or you could write valid JavaScript that works in Node LTS, which is already fairly modern and skip the transpilers. 🙂 

> Many libraries use [tsup](https://tsup.egoist.dev/) and I recognize Rslib plays the same role. I've started trying it, but the output looks better. However, the build time seems to be longer than Rslib.

Okay, you may not need tsup either

tats-u:
I republished packages using tsup.
It was very easy to configure without being suffered from tsconfig.json.
Could you examine the content of published packages again? I believe it got closer to what you want it to be.
Also all of my packages were able to get the \[TS] badge thanks to your advice.

- https://www.npmjs.com/package/micromark-extension-cjk-friendly-util?activeTab=code
- https://www.npmjs.com/package/micromark-extension-cjk-friendly?activeTab=code
- https://www.npmjs.com/package/micromark-extension-cjk-friendly-gfm-strikethrough?activeTab=code
- https://www.npmjs.com/package/remark-cjk-friendly?activeTab=code
- https://www.npmjs.com/package/remark-cjk-friendly-gfm-strikethrough?activeTab=code

If you prefer UNPKG,

- https://www.unpkg.com/browse/micromark-extension-cjk-friendly-util@1.1.0/dist/
- https://www.unpkg.com/browse/micromark-extension-cjk-friendly@1.1.0/dist/
- https://www.unpkg.com/browse/micromark-extension-cjk-friendly-gfm-strikethrough@1.1.0/dist/
- https://www.unpkg.com/browse/remark-cjk-friendly@1.1.0/dist/
- https://www.unpkg.com/browse/remark-cjk-friendly-cjk-strikethrough@1.1.0/dist/

> Or you could write valid JavaScript that works in Node LTS, which is already fairly modern and skip the transpilers. 🙂

I want transpilers to be an "insurance".

ChristianMurphy:
LGTM

> I want transpilers to be an "insurance".

I hear you 🙂 
You have a library version set, which will validate your source sticks to features in that version https://github.com/tats-u/markdown-cjk-friendly/blob/5a3c70679e2954d846517f93ee4376df8678b481/packages/remark-cjk-friendly/tsconfig.json#L3
Current LTS of Node supports es2022 https://github.com/remarkjs/remark/blob/06ac0b526d6504fcce5375e25ec461c56539b617/tsconfig.json#L9

tats-u:
Do you have anything else that you still concerned about?

doc/plugins.md
```diff
@@ -64,6 +64,9 @@ The list of plugins:
   (👉 **note**: alternative to [`remark-capitalize`](https://github.com/zeit/remark-capitalize))
 * 🟢 [`remark-cite`](https://github.com/benrbray/remark-cite)
   – new syntax for Pandoc-style citations
+* 🟢 [`remark-cjk-friendly` / `remark-cjk-friendly-gfm-strikethrough`](https://github.com/tats-u/markdown-cjk-friendly)
+  – mitigates the problem in CommonMark/GFM that emphasis or
+  strikethrough are sometimes not recognized as intended in CJK languages
```
wooorm, line 69: Please use separate entries: you are adding 2 plugins, so please add 2 bullets too.
You can use *deep* links into your monorepo so that each link points to the exact docs.

Also, can you use shorter descriptions? Try to explain to the reader why they should (not) use your plugin, as quickly as possible.

Thank you!

tats-u: I split the entry and made the description so short that `node --run format` complains nothing.

tats-u:
Is the nested entry OK? `-gfm-strikethrough` is just a subpackage of `remark-cjk-friendly`.

wooorm:
No, thank you. Please keep with the style that is in place. Please use *short* descriptions. In-depth information can go *in* your documentation.

wooorm (approved):

wooorm:
Thank you!

PR #1430, opened 2025-03-22
added remark-react-liquid-tag

alfredosalzillo:
Initial checklist
- [x] read the support docs
- [x] read the contributing guide
- [x] agree to follow the code of conduct
- [x] searched issues and discussions and couldn’t find anything or linked relevant results below
- [x] made sure the docs are up to date
- [x] included tests (or that’s not needed) (N/A)

Description of changes

Adds the following plugins to plugins.md:

- [remark-react-liquid-tag](https://github.com/alfredosalzillo/remark-react-liquid-tag)

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -361,6 +361,8 @@ The list of plugins:
   — new syntax for wiki links (rehype compatible)
 * 🟢 [`remark-yaml-config`](https://github.com/remarkjs/remark-yaml-config)
   — configure remark w/ YAML
+* 🟢 [`remark-react-liquid-tag`](https://github.com/alfredosalzillo/remark-react-liquid-tag)
+  — add support for custom embeds using liquid tag syntax
 
 <!--lint enable media-style-->
 


wooorm:
Hi there! This doesn’t seem to be a plugin? There is no `remarkLiquidTag` exposed? A bunch of the types are a bit weird, I recommend taking time with the existing plugins, and https://unifiedjs.com/learn/guide/create-a-remark-plugin/

For the idea, you likely should not integrate this closely with react and liquid.
The remark ecosystem is not solely for react.

I recommend looking at how `remark-directive` works.
You could expose *its* nodes.
Then, liquid will work with existing remark/rehype/unified tools.
This is for example also what https://github.com/incentro-ecx/remark-github-admonitions-to-directives does.

alfredosalzillo:
@wooorm The guide states that the plugin should be exported as default export, not a named export, the plugin exports the plugin function as default, like the example https://unifiedjs.com/learn/guide/create-a-remark-plugin/#plugin

wooorm:
Indeed, plugin use the default export!

alfredosalzillo:
@wooorm then I don't understand what do you mean with

> There is no remarkLiquidTag exposed?

wooorm:
I glossed over the code and found several things missing/wrong. My `?` indicates a question, uncertainty. I also posted several other things. It looks like your `remarkReactLiquidTag` ***returns*** a plugin, and thus is not a plugin.

PR #1439, opened 2025-04-16
Updated `remark-mermaidjs`

pastacolsugo:
The package repository has been archived by the author in favour of `rehype-mermaid`.

### Initial checklist

* [ ] I read the support docs 
* [ ] I read the contributing guide 
* [ ] I searched issues and discussions and couldn’t find anything or linked relevant results below 

### Description of changes

Changed plugin status in the list


JounQin (approved):

remcohaszing (changes requested):
The plugin is deprecated, not broken. I suggest we remove it from the list instead.

wooorm:
if it is no longer suggested for use, then it is better to remove it from the list of suggestions. So :+1: to Remco’s suggestion. @pastacolsugo, can you instead remove it?

PR #1457, opened 2025-08-24
Add `remark-arrow` to list of plugins

cocoliliace:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Add [`remark-arrow`](https://sr.ht/~liliace/remark-arrow) to list of plugins.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -51,6 +51,8 @@ The list of plugins:
   compatible)
 * 🟢 [`remark-api`](https://github.com/wooorm/remark-api)
   — generate an API section
+* 🟢 [`remark-arrow`](https://sr.ht/~liliace/remark-arrow)
+  — replace ASCII arrows with Unicode arrows
 * ⚠️ [`remark-attr`](https://github.com/arobase-che/remark-attr)
   — new syntax to add attributes to markdown
 * 🟢 [`remark-behead`](https://github.com/mrzmmr/remark-behead)


remcohaszing (approved):
Thanks! I like it!

ChristianMurphy (changes requested):
Thanks for sharing @cocoliliace !
The code looks good!

A few thoughts.
* consider adopting continuous integration/testing (CI/CD). I'm not as familiar with source hut, but it looks like it has its own hosted version at https://man.sr.ht/builds.sr.ht/
* Please also include a LICENSE (.txt,.md) in the repository with the license text.
* it looks like you are going for MPL 2.0, that license also requires a comment to be included in each source file https://www.mozilla.org/en-US/MPL/headers/

cocoliliace:
Hi Christian thank you for your feedback and your thoughtful links :D.

- Is CI/CD required for plugins to be listed here? I had considered adding it but decided not to at this time. It requires a paid account on sourcehut, and I have already set up a [git pre-commit hook](https://git-scm.com/book/ms/v2/Customizing-Git-Git-Hooks#_committing_workflow_hooks) to run the linter and tests locally before each commit.
- I had included the license text in the `COPYING` file. Looks like Github is able to detect the license from it if `LICENSE` is not found. Example: https://github.com/bminor/bash
- It looks like the header text is not strictly required, from [Exhibit A from MPL-2.0](https://www.mozilla.org/en-US/MPL/2.0/#exhibit-a---source-code-form-license-notice):
> If it is not possible or desirable to put the notice in a particular file, then You may include the notice in a location (such as a LICENSE file in a relevant directory) where a recipient would be likely to look for such a notice.

In my case it falls under the not desirable category, in my subjective view of what's desirable.

PR #1474, opened 2026-01-07
Add remark-validate-relative-links plugin

webpro:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Part of

- https://github.com/webpro/remark-preset-webpro
- https://github.com/webpro/mdxlint-preset-webpro

Has already been quite useful in two projects:

- https://github.com/release-it/release-it/commit/6e6dd4b893bd53a621ea2bee9ad48d5fa42f6279
- https://github.com/webpro-nl/knip/commit/a8d0bfd0fb95e8b9fb8ed5847093ba03e7364d5f

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -357,6 +357,8 @@ The list of plugins:
   — turn bolds, italics, and code into UTF 8 special characters
 * 🟢 [`remark-validate-links`](https://github.com/remarkjs/remark-validate-links)
   — check links to headings and files
+* 🟢 [`remark-validate-relative-links`](https://github.com/webpro/remark-validate-relative-links)
+  — check links in local files and heading
 * ⚠️ [`remark-variables`](https://github.com/mrzmmr/remark-variables)
   — new syntax for variables
 * 🟢 [`remark-vdom`](https://github.com/remarkjs/remark-vdom)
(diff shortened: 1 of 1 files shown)


doc/plugins.md
```diff
@@ -357,6 +357,8 @@ The list of plugins:
   — turn bolds, italics, and code into UTF 8 special characters
 * 🟢 [`remark-validate-links`](https://github.com/remarkjs/remark-validate-links)
   — check links to headings and files
+* 🟢 [`remark-validate-relative-links`](https://github.com/webpro/remark-validate-relative-links)
```
ChristianMurphy: I'm not sure I understand this plugin from the readme, it looks like it does the same thing as https://github.com/remarkjs/remark-validate-links ?

ALso this doesn't appear to have types? Or be published to npm? Or have a CI/CD job testing it. All are things we look for in something to recommend to the community

ChristianMurphy (changes requested):

wooorm:
> I think relative paths can’t start with a protocol, slash, or hashtag.

`//` is relative to the current protocol, also `?` starts query parameters.
But then `?` and `#` *are* relative to the current path. And `/asd` is relative to the repo. And the `//` is relative to the protocol. So “relative” is a very vague term.
Not the exact same but maybe useful are https://github.com/micromark/micromark/blob/774a70c6bae6dd94486d3385dbd9a0f14550b709/packages/micromark-util-sanitize-uri/dev/index.js#L27 and https://github.com/micromark/micromark/blob/774a70c6bae6dd94486d3385dbd9a0f14550b709/test/io/misc/dangerous-protocols.js

PR #1476, opened 2026-02-01
Add `remark-plugin-autonbsp` to list of plugins

denisinvader:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Add [remark-plugin-autonbsp](https://github.com/denisinvader/remark-plugin-autonbsp) to the list of plugins

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -258,7 +258,8 @@ The list of plugins:
   — new syntax for mentions w/ configurable existence check (new node
   type, rehype compatible)
 * 🟢 [`remark-plugin-autonbsp`](https://github.com/denisinvader/remark-plugin-autonbsp)
-  —  replace whitespace with non-breaking spaces to avoid hanging articles, prepositions and digits
+  —  replace whitespace with non-breaking spaces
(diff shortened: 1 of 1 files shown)


remcohaszing (approved):
Cool project!

A few tips:
- You can build the project with `tsc`. You don’t need Vite.
- If you do really want to use Vite for some reason: Vitest understands `vite.config.ts`, so you can remove `vitest.config.ts`.
- The `types` field describes how to resolve types for the `main` field. You didn’t specify the `main` field. You should specify it if you want to support it. If you do specify the `main` field, you no longer need the `types` field.
- For my personal projects I no longer support the `main` field. I find it more valuable teaching people to use the correct `module` settings the TS config.
- You can simplify this:
  ```jsonc
    "exports": {
      ".": "./dist/index.js"
    }
  ```
  into:
  ```jsonc
    "exports": "./dist/index.js"
  ```
- The `prepublishOnly` script runs before packing, but only if publishing directly. I recommend using `prepack`, which always happens before packing.
- Perhaps this would be better as a rehype plugin instead of a remark plugin.

PR #1481, opened 2026-03-26
ci: stop testing floating node alias

astrochicken-hatch:
### Initial checklist

* [x] I read the support docs

### Description of changes

Stop using the floating `node` alias in CI and test Node 24 explicitly in the main and canary jobs.

The current workflow resolves `node-version: node` to Node 25, which currently breaks this repo and the canary jobs inside `yargs` during test coverage runs. Pinning a stable supported release fixes the current CI failure without changing package code.

Evidence from failing PR run `22593660529`:
- `node-version: node` resolved to `v25.7.0`
- `npm test` failed with `ReferenceError: require is not defined in ES module scope` in `node_modules/yargs/yargs`
- both canary jobs hit the same Node 25 failure

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: node
+          node-version: 24
       - run: npm install
       - run: npm run build
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


ChristianMurphy (commented):
Testing latest is intentional, it catches issues when new releases some out.

astrochicken-hatch:
Makes sense, thanks for the clarification.

I dug into it locally and the failure might just be a Node 25.7.0 incompatibility in c8/yargs rather than a remark regression. Replacing `node` with `24` removes the intentional latest check, so this PR is the wrong fix for your CI setup.

Closing this out.

PR #1489, opened 2026-06-07
doc: add @webc.site/math-remark to plugins list

x-at-01:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This PR adds `@webc.site/math-remark` to the lists of remark plugins in `doc/plugins.md`.

- `@webc.site/math-remark` is a plugin that transforms math nodes (parsed by `remark-math`) into native browser-supported MathML Core elements at compile time.
- It is extremely small (around 3.58KB gzipped) and offloads layout rendering entirely to browser-native layout engines, saving significant JS/CSS bundle size on client-side math rendering.

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -231,6 +231,8 @@ The list of plugins:
   — serialize markdown as man pages (roff)
 * 🟢 [`remark-math`](https://github.com/remarkjs/remark-math)
   — new syntax for math (new node types, rehype compatible)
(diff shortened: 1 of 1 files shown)


remcohaszing (changes requested):
Thanks, but I don’t think this is the right place for it. You’re generating HTML. The correct approach would be a rehype plugin, not a remark plugin.

For an example, have a look at [rehype-katex](https://github.com/remarkjs/remark-math/blob/main/packages/rehype-katex/lib/index.js).

